PR #21, opened 2016-03-19
unbreak two examples

alon:
missing two casts of float64 to int and the initial call to queue.get
must not be blocking so push a silence initial sample.
plus a missing import of numpy as np.

### DIFF
--- a/examples/plot_input.py
+++ b/examples/plot_input.py
@@ -2,6 +2,7 @@
 """Plot the live microphone signal(s) with matplotlib."""
 import argparse
 from queue import Queue, Empty
+import numpy as np
 
 
 def int_or_str(text):
@@ -86,7 +87,7 @@ try:
         device_info = sd.query_devices(args.device, 'input')
         args.samplerate = device_info['default_samplerate']
 
-    length = np.ceil(args.window * args.samplerate / (1000 * args.downsample))
+    length = np.ceil(args.window * args.samplerate / (1000 * args.downsample)).astype(int)
     plotdata = np.zeros((length, len(args.channels)))
 
     fig, ax = plt.subplots()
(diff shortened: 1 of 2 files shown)


mgeier:
Thanks for scrutinizing my examples, I very much appreciate that!

> missing two casts of float64 to int

I'm not sure if `astype()` is the "nicest" solution here (even though I used it myself in `spectrogram.py`) ...
It makes sense for arrays, but I think it would be nicer to convert the Python `float` to a Python `int` without involving NumPy at all.

One alternative for `spectrogram.py` would be:

```python
import math
...
... = math.ceil(...)
... = math.floor(...)
```

In the case of `plot_input.py` it would probably be simpler to round towards zero (since there it's a somewhat arbitrary decision):

```python
length = int(...)
```

What do you think about those alternatives?

Apart from how it looks, there is also a nasty edge case that raises a nice error when using the Python stuff but just returns a crazy result when using NumPy:

```python
>>> np.ceil(float('inf')).astype(int)
-9223372036854775808
```

> the initial call to queue.get must not be blocking

Why is that?

> a missing import of numpy as np

It's not missing, it's just at a non-typical place.

My idea with that was to avoid printing a traceback if NumPy is not installed.
But I guess that's highly unlikely anyway, so we might as well move `import numpy as np` to the top of the file.

I would still keep the `sounddevice` import inside the `try` block, since it may raise an exception at import time, even if it is properly installed.

I'm not sure if the `matplotlib` imports should go to the top or stay where they are.

What are your thoughts about that?

mgeier:
@alon Any more comments on that?
I'd be especially interested in the `queue` thing ... but I'd also be interested in your opinion on the other points I mentioned above.

alon:
Hi,

 Sorry but this is completely out of my working set right now. Wrt the queue, I do not recall but I had to do that to get it working, sorry for being vague. I expect it was blocking like I wrote. Perhaps you used it differently, I'm not likely to get back to it any time soon - I ended up using gstreamer for my needs so I'm not working with your library after that initial bout.

 Wrt the type conversions, I agree it is nicer to avoid astype, your logic sounds good (using ceil/floor).

Alon

mgeier:
I fixed the rounding in fd9f706b90d5c64d9c7a6903e62724c251da012f.

If there are still problems with the queue, please re-open this or create a new issue.

alon:
I appreciate it, thank you.

PR #31, opened 2016-07-31
Convert CFFI strings to native Python strings

mgeier:
i.e. Unicode strings in Python 3 and bytes in Python 2.


mgeier:
Merged #32 instead.

Issue #4, opened 2015-07-23
Enable platform-specific features from PortAudio


mgeier:
See https://github.com/bastibe/PySoundCard/issues/16.

mgeier:
One possible feature would be the "exclusive mode" of WASAPI, see http://stackoverflow.com/questions/31306010/how-to-enable-wasapi-exclusive-mode-in-pyaudio

mgeier:
`PaAlsa_EnableRealtimeScheduling` might be interesting for ALSA: http://portaudio.com/docs/v19-doxydocs-dev/pa__linux__alsa_8h.html#accc41e7fd89dc0a6163d54a4f25fb242

raecke:
It would be interesting to be able to select arbitrary channels for input and output. On Windows this would is needed with ASIO. If you currently want to send data to output channels 9 and 10, you have to send data to all channels below 11. This means that sharing the Soundcard with other applications becomes impossible.
The hostapis WDM-KS, MME, Directsound, normally only present 2 channels so it is not a problem there.

mgeier:
@raecke I created an experimental implementation in #34. Can you please check if it works for you and leave a comment there?

mgeier:
There is also `PaAlsa_SetNumPeriods`, which is by default set to `4`, which seems random:

> By default the number of periods is 4, this is the lowest number of periods that works well on the author's soundcard. 

http://portaudio.com/docs/v19-doxydocs-dev/pa__linux__alsa_8h.html#a76f60d6f1ba6a298c038a9f3a6dccc53

It looks like this has to be called before opening a stream.

mgeier:
OK, I'm trying to make an exhaustive list of the current host-API-specific settings, let's see if that works:

`hostApiSpecificStreamInfo` to be specified separately for the input and output device when creating a stream:

* [PaAlsaStreamInfo::deviceString](http://portaudio.com/docs/v19-doxydocs-dev/structPaAlsaStreamInfo.html#a442fbbb635cda0cece4e7af397d453f9), only works if device index is set to `paUseHostApiSpecificDeviceSpecification` (a.k.a. `-2`)
* [PaAsioStreamInfo::channelSelectors](http://portaudio.com/docs/v19-doxydocs-dev/structPaAsioStreamInfo.html#a2a04b8c7166044c009af7fdeca754c4d), needs `PaAsioStreamInfo::flags` to contain `paAsioUseChannelSelectors`
[...]

PR #34, opened 2016-08-09
Add extra_settings and AsioSettings

mgeier:
WARNING: this is untested!

See #4.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -22,7 +22,7 @@ API Documentation
    :exclude-members: RawInputStream, RawOutputStream, RawStream,
                      InputStream, OutputStream, Stream,
                      CallbackFlags, CallbackStop, CallbackAbort,
-                     PortAudioError, DeviceList
+                     PortAudioError, DeviceList, AsioSettings
 
 .. autoclass:: Stream
    :members:
@@ -50,6 +50,8 @@ API Documentation
 .. autoclass:: CallbackAbort
 
 .. autoclass:: PortAudioError
+
+.. autoclass:: AsioSettings
 
 .. only:: html
 
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -224,6 +224,19 @@ signed long Pa_GetStreamWriteAvailable( PaStream* stream );
 /* not implemented: Pa_GetStreamHostApiType */
 PaError Pa_GetSampleSize( PaSampleFormat format );
 void Pa_Sleep( long msec );
+
+/* pa_asio.h */
+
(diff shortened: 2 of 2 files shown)


mgeier:
As long as there are no docs available, this is one way to use it:

```python
import sounddevice as sd
extra = sd.AsioSettings(channel_selectors=[9, 10])
sd.play(mydata, fs, extra_settings=extra)
```

raecke:
This was unexpectedly fast.

I tested recording and playing with an arbitrary channel selection with a 10 channel soundcard. It works perfectly.

mgeier:
Thanks for testing!

Can you please do further tests:

* [ ] Using a non-existing channel number in `channel_selectors` should raise an error. Can you please post the error here? Please also try `0` and a negative number.

* [ ] If `len(channel_selectors) != channels` the library might crash. Can you please try this with `channel_selectors` longer and shorter than the number of available hardware channels? Can you please artificially reduce the number of channels with `channels=N` and then try it with a length between `N` and the actual number of hardware channels?

* [ ] What happens if the length is zero?

raecke:
1.   If any channel number is invalid (`0` is a valid channel number.), I get the error message :    
    `PortAudioError: Error opening InputStream: Invalid number of channels`.

2.  If the length of the `channel_selectors` list greater than the `channels` parameter, only the first part of the list is used. if the length of the `channel_selectors` list is smaller, I get the error :     
    `PortAudioError: Error opening InputStream: Invalid number of channels`.

3. If the length of the `channel_selectors` list zero. I get the error message:  
    `PortAudioError: Error opening InputStream: Invalid number of channels`.

mgeier:
Thanks for testing!

All those sound like reasonable error behaviors, right?

Are the channel numbers zero-based?

Anything else I should add to the docs?

raecke:
- The first error message is not ideal because you would think the length of the `channel_selectors` list is wrong, but it is the content that is wrong. But as this is the error that portaudio reports, I would be reluctant to change this.
- The channel numbers are zero based.
- I think the documentation needs:
 - The example above.
 - The remark that the channel numbers are zero based.
 - The remark that the information about the number of channels still needs to be provided and has to be equal to `len(channel_selectors)`.

mgeier:
Thanks! I'll make the suggested changes next week.

mgeier:
I've added the suggested information to the docs: http://python-sounddevice.readthedocs.io/en/latest/#sounddevice.AsioSettings

If there is something unclear/missing, or if the examples don't work as shown, please make a PR with the corrections!

Issue #69, opened 2017-01-20
"Unknown Error" When Stopping a Blocking Play (Mac OS X)


riggsd:
I'm using sounddevice 0.3.6 (which magically pulled in PortAudio V19.6.0-devel) on Mac OS X 10.9.5 with python.org Python 2.7.12.

When I `stop()` a blocking `play()` from another thread, I see the following message on stderr:

    ||PaMacCore (AUHAL)|| Error on line 2754: err='-50', msg=Unknown Error

The following example code triggers this error:

```python
import time
from threading import Thread
from scipy.io import wavfile
import sounddevice


class AudioThread(Thread):
    def __init__(self, fname):
        Thread.__init__(self)
        self.fname = fname

    def run(self):
        samplerate, signal = wavfile.read(self.fname)
        sounddevice.play(signal, samplerate, blocking=True)

    def stop(self):
        sounddevice.stop()


thread = AudioThread('test.wav')
thread.start()
time.sleep(1)
thread.stop()
```

Calling `stop()` on a *non-blocking* `play()` from the *same* thread does not produce an error.

mgeier:
PortAudio is not claiming to be thread-safe: https://app.assembla.com/wiki/show/portaudio/Tips_Threading

Since opening and closing a stream doesn't work for you, you can try writing your own callback function and implement the "stop" functionality without actually stopping the stream (but just writing zeros afterwards).

Or you can try to somehow have `play()` and `stop()` on the same thread.

Often creating a separate thread isn't even necessary, since the callback function is called in a dedicated thread (created by PortAudio) anyway.

riggsd:
I am never creating multiple simultaneous threads, my application has a GUI thread and either one (if currently playing a sound) or zero (if not playing) audio thread. Both `play()` and `stop()` always happen in the same thread. This doesn't seem to be a multithreading issue as described in the above PortAudio documentation, but I most certainly don't understand the internals of PortAudio as well as you do.

I'd be fine with calling `play()` asynchronously if I could ask for the current status as to whether it is still playing audio or not. Is there a way to do this in the API? If I could query whether playback was still in progress, I wouldn't need the extra thread. My example code doesn't show it, but by asking whether `Thread.is_alive()` I can tell whether the blocking `play()` is still playing or not.

mgeier:
But in your code example, `play()` and `stop()` are called from different threads ... is your real use case different than that?

[...]

PR #70, opened 2017-01-24
Add `get_stream()` function for accessing the current Stream

riggsd:
This PR adds a top-level function `get_stream()` for use with the imperative-style API of `play()`, `rec()`, `playrec()`, and `stop()`. It allows end users to access the current state of playback and recording at the Stream level.

This PR effectively closes Issue #69.

### DIFF
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -600,20 +600,6 @@ def get_stream():
         the last invocation of `play()`, `rec()` or `playrec()`
         respectively.
 
-    Examples
-    --------
-
-    >>> # Display progress indicator while audio is playing
-    >>> import sounddevice as sd
-    >>> from time import sleep
-    >>> sd.play(...)
-    >>> print('Playing sound...', end='')
-    >>> stream = sd.get_stream()
-    >>> while stream.active:
-    >>>     sleep(1.0)
-    >>>     print('.', end='')
-    >>> print(' Done.')
-
     """
     if _last_callback:
         return _last_callback.stream
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -587,6 +587,38 @@ def get_status():
         raise RuntimeError('play()/rec()/playrec() was not called yet')
 
 
+def get_stream():
+    """Get a reference to the current Stream.
+
+    This applies only to Streams created by calls to `play()`, `rec()`,
+    or `playrec()`.
+
+    Returns
+    -------
+    Stream
+        An `OutputStream`, `InputStream`, or `Stream` associated with
+        the last invocation of `play()`, `rec()` or `playrec()`
+        respectively.
+
+    Examples
+    --------
+
+    >>> import sounddevice as sd
+    >>> from time import sleep
+    >>> sd.play('test.wav')
+    >>> print('Playing test.wav...', end='')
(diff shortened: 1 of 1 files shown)


mgeier:
Thanks, that's great!

The only think I don't like is the example.
I don't want to put this Python 2 work-around for `play(..., blocking=True)` into the documentation.
If you want, you can make a separate section in the docs just for Python 2 compatibility where such a work-around would make sense.

Another - minor - point is that it doesn't work like this, because `play()` taking a file name argument is wishful thinking.

What about using something simpler, like e.g. the following?

```pycon
>>> import sounddevice as sd
>>> sd.play(myarray, 48000)
>>> stream = sd.get_stream()
>>> stream.active
True
>>> sd.stop()
>>> stream.active
False
```

(It's also untested ...)

sounddevice.py
```diff
@@ -587,6 +587,38 @@ def get_status():
         raise RuntimeError('play()/rec()/playrec() was not called yet')
 
 
+def get_stream():
+    """Get a reference to the current Stream.
```
mgeier: What is the reason to capitalize "Stream" (also in the next line)?

I'm just asking because I'm not a native speaker ...

sounddevice.py
```diff
@@ -587,6 +587,38 @@ def get_status():
...
+    >>> import sounddevice as sd
+    >>> from time import sleep
+    >>> sd.play('test.wav')
+    >>> print('Playing test.wav...', end='')
+    >>> stream = sd.get_stream()
+    >>> while stream.active:
+    >>>     sleep(1.0)
+    >>>     print('.', end='')
+    >>> print(' Done.')
+    """
```
mgeier: For consistency, the docstring should end with a blank line.
I think this is the NumPy documentation style ...

sounddevice.py
riggsd: I was referring to the fact that a `Stream` is a concrete class, but this is inconsistent with existing docstrings. Fixed.

sounddevice.py
riggsd: Fixed.

riggsd:
@mgeier, the example code has nothing to do with Python 2 workarounds! It's an example of displaying a progress indicator during long-running playback.

Apologies for the pseudocode filename, I had inserted it as placeholder because I didn't want to imbed `scipy.io.wavfile.read()` into a docstring. For consistency with other docstrings, I've fixed the example to use a placeholder as you've done elsewhere:

```python
>>> sd.play(...)
```

mgeier:
> the example code has nothing to do with Python 2 workarounds! It's an example of displaying a progress indicator during long-running playback.

OK, you are right, it's not exactly the work-around, but it's damn close ...

But apart from that, I also think that the code is a bit too complicated to type into the interactive interpreter (and that's what the doctest `>>>`'s suggest).
Also, the real prompt will actually look different for the lines in the `while` loop.

I would prefer something that can be reasonably typed into an interactive prompt.

Your example would probably fit better to the other code examples in the README?

riggsd:
I have removed the example from the `get_stream()` docstring.

mgeier:
@riggsd Thanks!

PR #82, opened 2017-04-04
adds error codes to PortAudioErrors and a new HostAPIError type

tgarc:
Attach portaudio error codes to portaudio errors to allow for better error handling. This would allow users to handle specific portaudio and host api error codes. The `_check` function was also updated to use these new exceptions. The `from_errno` and `from_errinfo` are a convenience that I thought might make it easier for users to raise an exception with a default message. Hopefully this doesn't add too much complexity to the code.

### DIFF
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -544,7 +544,7 @@ def playrec(data, samplerate=None, channels=None, dtype=None,
     input_frames = ctx.check_out(out, output_frames, channels, dtype,
                                  input_mapping)
     if input_frames != output_frames:
-        raise PortAudioError('len(data) != len(out)')
+        raise ValueError('len(data) != len(out)')
     ctx.frames = input_frames
 
     def callback(indata, outdata, frames, time, status):
@@ -951,7 +951,7 @@ class _StreamBase(object):
             self._channels = iparameters.channelCount, oparameters.channelCount
             self._samplesize = isize, osize
             if isamplerate != osamplerate:
-                raise PortAudioError(
+                raise ValueError(
                     'Input and output device must have the same samplerate')
             else:
                 samplerate = isamplerate
@@ -2289,13 +2289,39 @@ class default(object):
         vars(self).clear()
         self.__init__()
 
+
 if not hasattr(_ffi, 'I_AM_FAKE'):
     # This object shadows the 'default' class, except when building the docs.
     default = default()
 
 
 class PortAudioError(Exception):
-    """This exception will be raised on PortAudio errors."""
+    """This exception will be raised on PortAudio errors.
+
+    Attributes
+    ----------
+    args
+        A variable length tuple containing the following elements when
+        available:
+
+        1) A string describing the error
+        2) The PortAudio ``PaErrorCode`` value
+        3) A 3-tuple containing the host API index, host error code, and the
+           host error message (which may be an empty string)
+
+    """
+
+    def __str__(self):
+        errormsg = self.args[0] if self.args else ''
+        if len(self.args) > 1:
+            errormsg = "{0} [PaErrorCode {1}]".format(errormsg, self.args[1])
+        if len(self.args) > 2:
+            host_api, hosterror_code, hosterror_text = self.args[2]
+            hostname = query_hostapis(host_api)['name']
+            errormsg = "{0}: '{1}' [{2} error {3}]".format(
+                errormsg, hosterror_text, hostname, hosterror_code)
+
+        return errormsg
 
 
 class CallbackStop(Exception):
@@ -2776,19 +2802,27 @@ def _split(value):
     return invalue, outvalue
 
 
-def _check(err, msg=""):
+def _check(err, msg=''):
     """Raise error for non-zero error codes."""
-    if err < 0:
-        msg += ': ' if msg else ''
-        if err == _lib.paUnanticipatedHostError:
-            info = _lib.Pa_GetLastHostErrorInfo()
-            hostapi = _lib.Pa_HostApiTypeIdToHostApiIndex(info.hostApiType)
-            msg += 'Unanticipated host API {0} error {1}: {2!r}'.format(
-                hostapi, info.errorCode, _ffi.string(info.errorText).decode())
-        else:
-            msg += _ffi.string(_lib.Pa_GetErrorText(err)).decode()
-        raise PortAudioError(msg)
-    return err
+    if err >= 0:
+        return err
+
+    errormsg = _ffi.string(_lib.Pa_GetErrorText(err)).decode()
+    if msg:
+        errormsg = "{0}: {1}".format(msg, errormsg)
+
+    if err == _lib.paUnanticipatedHostError:
+        # (gh82) We grab the host error info here rather than inside
+        # PortAudioError since _check should only ever be called after a
+        # failing API function call. This way we can avoid any potential issues
+        # in scenarios where multiple APIs are being used simultaneously.
+        info = _lib.Pa_GetLastHostErrorInfo()
+        host_api = _lib.Pa_HostApiTypeIdToHostApiIndex(info.hostApiType)
+        hosterror_text = _ffi.string(info.errorText).decode()
+        hosterror_info = host_api, info.errorCode, hosterror_text
+        raise PortAudioError(errormsg, err, hosterror_info)
+
+    raise PortAudioError(errormsg, err)
 
 
 def _get_device_id(id_or_query_string, kind, raise_on_error=False):


mgeier:
Thanks for this PR!

Is it really necessary to create a new exception class?
Would anybody actually `try`/`catch` for this?

TBH, I don't have a lot of experience with creating new exception classes, are there any guidelines we should follow?

I saw in the Python docs that there is this `args` attribute, should we use that?
AFAICT, it is excessively used in [OSError](https://docs.python.org/3/library/exceptions.html#OSError) and subclasses, which actually sounds like a quite similar situation.

What about using

```python
args = pa_error_code, error_text
```

... in the general case, and for the special case `paUnanticipatedHostError` we can use

```python
args = pa_error_code, error_text, host_api, host_error_code, host_error_text
```

?

This should then be easy to handle programmatically, right?

I also don't know why we would need `from_errno()` and `from_errinfo()`, can't this just be all done in a single constructor?

Something like this:

```python
def _check(err, msg=''):
    if err < 0:
        raise PortAudioError(err, msg)
    return err
```

Everything else could happen inside of `PortAudioError.__init__()` ...

tgarc:
> Is it really necessary to create a new exception class?

TBH, I don't know how much of a use case the HostAPIError would have in the long run but I thought it might be useful for users implementing features that they expect could raise exceptions with certain audio api's. And logically, HostAPIError is more like an OSError, where as PortAudioError is an error directly related to the library so it makes sense that it's a separate class. I'm not dead set on adding this class though, as long as we can add the `hostapi`/`hosterrno` attributes to the PortAudioError class I'm fine with it.

> AFAICT, it is excessively used in OSError and subclasses, which actually sounds like a quite similar situation.

That was the idea, I was trying to make it look and feel like an OSError.

> I also don't know why we would need from_errno() and from_errinfo(), can't this just be all done in a single constructor?

The trouble I found is that I wanted to expose a simple way to use the portaudio utilities to build a useful default error message. Otherwise a user has to directly access the `_lib` object and call a series of functions to get the host error message. I guess your assumption is that only advanced users would want to raise their own PortAudioError exceptions and therefore those users won't mind using the portaudio API directly - is that right? If we want to make that assumption then what you propose would certainly be the easiest/simplest to implement.

One easy adaptation I could make was that the constructor *always* require an error number (or PaHostErrorInfo in case of HostAPIError) - that would elminate the need for `from_errno` and `from_errinfo`.

mgeier:
I can't imagine a situation where the user would want to raise a `PortAudioError`, can you please describe the use case(s) you are thinking about?

tgarc:
well...point taken. I can't really think of any that don't involve the user accessing the `_lib` directly trying to extend functionality.

I'd like to make some amendments. Do you have any issue with adding `hosterrno`/`hosterrmsg`/`hostapi` attributes to the `PortAudioError` class?

mgeier:
Thanks for your changes, now it looks much less confusing IMHO.

Sadly, it doesn't work yet. There are a few instances of `raise PortAudioError(...)` that are called without error number.
I think the first of them can (should?) be replaced by a `ValueError`. Probably also the fourth?
But the others are either caused by `NULL` pointers or by a `0` return value, and I think they should logically be "PortAudio errors". Or do you think there should be a second exception class after all?
Probably the error number should be `None` in this case?

> Do you have any issue with adding `hosterrno`/`hosterrmsg`/`hostapi` attributes to the `PortAudioError` class?

No.
I'm not sure if the names are good, though.
Although I acknowledge the similarity to `OSError`, I wouldn't go too far with it. Especially "errno" has already a meaning which IMHO we shouldn't abuse.

As I said, I don't have a lot of experience with creating exception classes, but I found this: https://julien.danjou.info/blog/2016/python-exceptions-guide
I think this sounds reasonable, and we could implement our exception class similar to the example in the section "Defining your exceptions properly".

I especially like the idea of passing only a single string argument to the parent constructor, because it avoids having to implement the `__str__()` method.

There is one aspect where I think I don't agree: I wouldn't create exception classes for all `sounddevice` errors, and I wouldn't derive some of them from e.g. `ValueError`. I think I would rather raise `ValueError` and friends directly.
But I guess my opinion on this is still malleable ...

Initially, I thought we should use a tuple for the `args` attribute, but I somewhat changed my mind.
I think it is still a possibility, but we shouldn't use a tuple *and at the same time* provide individual attributes for the tuple elements. There should be one obvious way, so it's either one or the other, but not both.

tgarc:
> There are a few instances of raise PortAudioError(...) that are called without error number...others are either caused by NULL pointers or by a 0 return value, and I think they should logically be "PortAudio errors". 

Yeah I think the cases you describe are legitimate portaudio errors that don't have an error code so I guess we can't always require one.

> Or do you think there should be a second exception class after all? Probably the error number should be None in this case?

No and yes. I don't think these errors justify another class, just an unfortunate corner case.

> I think the first of them can (should?) be replaced by a ValueError.

The logic that would cause this [line](https://github.com/tgarc/python-sounddevice/blob/f435854bbdd6cde41c3eca0b790a29405d6e7fcc/sounddevice.py#L520) to raise an error is a bit hairy but from a glance it doesn't look like portaudio itself would cause this condition. This does seem like it should be a ValueError.

> Probably also the fourth?

Agreed.

> I especially like the idea of passing only a single string argument to the parent constructor, because it avoids having to implement the __str__() method.

That's a simple enough change. I'll go ahead and make it.

To handle the cases where there is no associated portaudio error code, what do you think of changing the __init__ signature to: `__init__(self, msg=None, errorCode=None)`. That would make it so that you can still raise a portaudio error with just a string message or just an error code. Then the PortAudioError class can still take care of setting the appropriate attributes and generating additional error information when `errorCode` is not None.

mgeier:
> I also added a PaErrorCode class

This looks interesting, but how is it supposed to be used?

I like the new constructor signature, but probably the message argument could be obligatory?
In the implementation of `_check()` we could just set `msg=''` as default value. Or is this a silly idea?

sounddevice.py
```diff
@@ -2265,10 +2265,66 @@ def reset(self):
...
+                apiinfo = _lib.Pa_GetHostApiInfo(hostapi)
+                self.hosterrcode = info.errorCode
+                self.hostapi = _ffi.string(apiinfo.name).decode() if apiinfo else 'unknown'
+                self.hosterrmsg = _ffi.string(info.errorText).decode()
+                errmsg += "Unanticipated {0!r} host error".format(self.hostapi)
+            else:
+                errmsg += _ffi.string(_lib.Pa_GetErrorText(errcode)).decode()
+            if msg is not None:
+                errmsg += ": {0!s}".format(msg)
+            elif self.hosterrmsg:
```
mgeier: Why are you ignoring the error message if a `msg` is given?

sounddevice.py
```diff
@@ -2265,10 +2265,66 @@ def reset(self):
...
+                hostapi = _lib.Pa_HostApiTypeIdToHostApiIndex(info.hostApiType)
+                apiinfo = _lib.Pa_GetHostApiInfo(hostapi)
+                self.hosterrcode = info.errorCode
+                self.hostapi = _ffi.string(apiinfo.name).decode() if apiinfo else 'unknown'
+                self.hosterrmsg = _ffi.string(info.errorText).decode()
+                errmsg += "Unanticipated {0!r} host error".format(self.hostapi)
+            else:
+                errmsg += _ffi.string(_lib.Pa_GetErrorText(errcode)).decode()
+            if msg is not None:
+                errmsg += ": {0!s}".format(msg)
```
mgeier: Why do you need a `!s` here?

sounddevice.py
```diff
@@ -2265,10 +2265,66 @@ def reset(self):
...
 
 class PortAudioError(Exception):
-    """This exception will be raised on PortAudio errors."""
+    """This exception will be raised on PortAudio errors.
+
+    Parameters
+    ----------
+    msg : optional
+        User error message.
+    errcode : int or None, optional
```
mgeier: This should be `int, optional`.

Do we even need to document the constructor arguments?
Would a user really want to create a `PortAudioError`?

mgeier:
About the attribute names:

I'd like to avoid unnecessary abbreviations in the public API. I guess "PA" is OK, but "err" and "msg" seem unnecessarily abbreviated to me.

I've already made a few suggestions [above](https://github.com/spatialaudio/python-sounddevice/pull/82#issuecomment-291880038), what about those?

I guess we can use `hostapi` instead of `host_api`, since former would be consistent with the already existing API. The use of "text" instead of "message" is inspired by the original PortAudio API, but I don't insist on it if you prefer "message".

sounddevice.py
```diff
@@ -2265,10 +2265,66 @@ def reset(self):
...
+    - The `hostapi`, `hosterrcode`, and `hosterrmsg` attributes will be
+      None except for the case when `paerrcode` ==
+      `paUnanticipatedHostError`.
+    
+    Attributes
+    ----------
+    paerrcode : int or None
+        The PaErrorCode integer value or None if there is no associated
+        error code.
+    hostapi : str or None
```
mgeier: I think it would make more sense to expose the host API number instead of its name.
`query_hostapis()` expects the number, the name doesn't really help that much, programmatically.
The error message should of course keep using the name.

sounddevice.py
tgarc: You're suggesting I should always print out the host error message? I guess that's more consistent since the portaudio error is always getting printed. I just did it to limit the size of the error message. Maybe it would make more sense to do:
```
if msg is None: 
    errmsg = [PaErrorCode <code>] portaudio error message 
    if hosterrmsg:
        errmsg += ': <hosterrmsg>'
elif paerrcode is not None: 
    errmsg =  [PaErrorCode <code>] msg
else:
    errmsg = msg
```
(where the part in curly brackets is optional)

sounddevice.py
tgarc: Actually, it's not necessary.

sounddevice.py
tgarc: probably documenting of the constructor arguments is unnecessary. I could remove them. The attribute documentation OTOH I think is useful for users.

sounddevice.py
tgarc: Yeah I would agree with that. I'll make that change.

tgarc:
> > I also added a PaErrorCode class

> This looks interesting, but how is it supposed to be used?

Like this:

```
try:
    sd.some_func()
except sd.PortAudioError as err:
    if err.paerrcode == sd.PaErrorCode.paNotInitialized:
        raise Exception("You dumb idiot, why didn't you initialize?")
    raise
```

It may be a little more verbose but it encapsulates all the error codes under one variable in the sounddevice namespace (instead of dumping all the error codes into the main namespace).

> I like the new constructor signature, but probably the message argument could be obligatory?

I would think the default portaudio error message would be informative enough in many cases. Why would you make it required?

> In the implementation of _check() we could just set msg='' as default value. Or is this a silly idea?

Having the default ```msg=None`` let's me disambiguate the lack of an error message from just an empty string. This would only be an important distinction if you want the msg to override the default portaudio message. See my comment [here](https://github.com/spatialaudio/python-sounddevice/pull/82/files/57db7a3f9f789d804c99780562e27ee066fd5145#r110758619).

sounddevice.py
mgeier: I wouldn't ignore any of the error messages, ever.
The separate message I'm providing in the `sounddevice` module is not meant to replace the PortAudio error message, it's rather meant to complement it by providing some high-level context.

E.g. my message would be "Error starting stream", while the PortAudio message would hopefully specify some details about what exactly went wrong.

mgeier:
Hmm ... it looks like you are implementing your own exception mechanism ... wouldn't something like this make more sense:

```python
try:
    sd.some_func()
except sd.PortAudioNotInitialized:
    raise Exception("You dumb idiot, why didn't you initialize?")
except sd.PortAudioError:
    raise Exception("whatever")
```

>> I like the new constructor signature, but probably the message argument could be obligatory?
>
> I would think the default portaudio error message would be informative enough in many cases. Why would you make it required?

No reason, just a feeling.
It doesn't really matter, since I guess we don't consider the constructor part of the public API, right?

sounddevice.py
mgeier: Definitely. I agree.

tgarc:
> Hmm ... it looks like you are implementing your own exception mechanism
But what you're suggesting would require adding a separate exception for each portaudio error code. I thought we were trying to avoid this?

mgeier:
> This is the same way you would catch a particular exception from e.g. the builtin IOError exception which similarly has an `errno` attribute.

But they also have their own exception classes.
I'm quite sure that it's more Pythonic to catch a certain exception class like `FileExistsError` instead of checking for the error number `EEXIST`, right?

> I thought we were trying to avoid this?

I don't know. I'm just asking questions ...

One important question might be: how many people will actually use this?

What about providing the "raw" error numbers and leaving it up to the users to find out on their own what those numbers mean?

I think we should either go for full support (i.e. Pythonic exception classes) or minimal support ("raw" error numbers). Your suggested enum class is somewhere in between, it is too fancy to be a "casual" feature, yet it is too awkward to be used in an idiomatic way.

Another problem with both your enum class and some exception sub-classes (if they are generated on the fly) is that it will be hard to document them properly. And if we have no documentation, we might as well just not have those features at all.

But still, I'm just thinking out loud ...

tgarc:
Let me take a step back here. The truth is that IMO if you're going to create a public Exception type, then it should be possible for a user to catch, reraise, or possibly even reraise as a new exception instance with a different error message. IOW, yes I think the user should be able to raise PortAudioError if for no other reason than because it is part of the public API. The host error case adds complications to implementing this and I don't think I could really justify it with any actual use cases at this point so I think I will remove it. (Of course `_check` will still include host error messages in the exceptions it raises). Instead, I think it would be more sensible to implement the PortAudioError constructor like this:

```
PortAudioError([msg])
PortAudioError(PaErrorCode, msg)
```

This is again similar to `IOError` but avoids the complications of a HostError case. How does this sound?

> Another problem with both your enum class and some exception sub-classes (if they are generated on the fly) is that it will be hard to document them properly

Agreed. The PaErrorCode is a bit clunky, I'll just take it out and we'll revisit it later if we see that people need it. For now we'll just provide raw error codes and leave it to the user for interpretation.

tgarc:
I think this implementation finally makes sense. Note that I didn't document the PortAudioError constructor, partly because I'm not sure what the right way to document a polymorphic constructor is. Do you think it needs it?

Also, are you okay with the constructor being polymorphic? I think it makes it more similar to the way an IOError works.

mgeier:
Thanks for the changes. I think it's a good idea to do the first part now and add the other features later, if needed.

I don't think the constructor arguments need documentation.

I'm not in general against the polymorphic constructor, but is it really necessary?
What about just having the message always as first argument?

> I think it makes it more similar to the way an IOError works.

Do you mean `OSError`?

This expects the error code as first argument, but I think it's not too crazy if we expect our error code as second argument, right?

sounddevice.py
```diff
@@ -2266,9 +2266,27 @@ def reset(self):
...
     default = default()
 
-
 class PortAudioError(Exception):
-    """This exception will be raised on PortAudio errors."""
+    """This exception will be raised on PortAudio errors. 
+
+    Attributes
+    ----------
+    paerrorcode : int or None
```
mgeier: Wouldn't `code` be enough?
`PortAudioError.paerrorcode` seems to be repeating itself.

sounddevice.py
```diff
@@ -2266,9 +2266,27 @@ def reset(self):
...
+        The PaErrorCode integer value or None if there is no associated error
+        code. Note that some PortAudioError's do not have an associated
+        PaErrorCode. In such a case `paerrorcode` will be None.
```
mgeier: Speaking of repetition, don't those sentences say the same thing twice?

sounddevice.py
```diff
@@ -2266,9 +2266,27 @@ def reset(self):
...
+    """
+    def __init__(self, *args):
+        if len(args) > 2:
+            raise TypeError("PortAudioError expected at most 2 arguments, got %d" % len(args))
+        elif len(args) == 2:
+            self.paerrorcode, errortext = args
+            errortext = "[PaErrorCode {0}] {1}".format(self.paerrorcode, errortext)
```
mgeier: Here the low-level error number comes before the high-level error text (for which it is actually meaningless).
I think I would prefer the `[PaErrorCode XYZ]` coming at the very end, after the low-level error message. Also, casual users will not be very interested in the error code, they should see the friendlier error message first.

sounddevice.py
```diff
@@ -2266,9 +2266,27 @@ def reset(self):
     # This object shadows the 'default' class, except when building the docs.
     default = default()
 
-
```
mgeier: This empty line was on purpose: [E302](http://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes).

sounddevice.py
```diff
@@ -2657,20 +2675,26 @@ def _split(value):
...
+        info = _lib.Pa_GetLastHostErrorInfo()
+        hostapi = _lib.Pa_HostApiTypeIdToHostApiIndex(info.hostApiType)
+        apiinfo = _lib.Pa_GetHostApiInfo(hostapi)
+        apiname = _ffi.string(apiinfo.name).decode() if apiinfo else 'unknown'
+        hosterrmsg = _ffi.string(info.errorText).decode()
+        msg += "Unanticipated {0} host error {1}".format(apiname, info.errorCode)
+        if hosterrmsg:
+            msg += ": '{0}'".format(hosterrmsg)
+    else:
+        msg += _ffi.string(_lib.Pa_GetErrorText(err)).decode()
```
mgeier: I guess both are possible, but is it better to obtain the error details here or move it to the `PortAudioError` constructor?

I'm kind of leaning to the latter ...

sounddevice.py
tgarc: yes, yes they do. I'll fix that.

sounddevice.py
tgarc: yup, missed that

sounddevice.py
tgarc: I was leaning towards doing it my way because it doesn't really make sense to me to have an exception class that forces a specific error message formatting. If I want to `raise PortAudioError("This is my error message. No one elses. Mine.")` I would expect to get my error message verbatim. I don't have much experience creating Exception classes either, but I've never seen an exception that forces formatting of error messages. Hence, _check serves as the default formatter of portaudio error messages.

This is just my feeling though, ultimately I'll it up to you what you want to implement.

sounddevice.py
tgarc: I agree, the error messages are printed in the wrong order.

tgarc:
Sorry, I killed the last commit with my latest commit. Anyway, I think this commit is closer to what we both wanted. You can ignore my [last comment](https://github.com/spatialaudio/python-sounddevice/pull/82#discussion_r111669342) since I basically contradicted my self with this commit. I'm happy with this now, let me know if you want any additional changes.

mgeier:
Thanks, looking good, now I can start nitpicking ...

sounddevice.py
```diff
@@ -2262,13 +2262,68 @@ def reset(self):
...
+        User error message.
+    code : int, optional
+        PaErrorCode value or None if there is no PaErrorCode associated
+        with the error.
+
+    Notes
+    -----
+    - The `host_api`, `host_code`, and `host_text` attributes will be None
+      except for the case when `code` == `paUnanticipatedHostError`.
+
```
mgeier: W293 blank line contains whitespace

sounddevice.py
```diff
@@ -2262,13 +2262,68 @@ def reset(self):
...
-    """This exception will be raised on PortAudio errors."""
+    """This exception will be raised on PortAudio errors.
+
+    Parameters
+    ----------
+    msg : str, optional
+        User error message.
+    code : int, optional
+        PaErrorCode value or None if there is no PaErrorCode associated
+        with the error.
```
mgeier: I think we don't need to document the constructor parameters, do we?

sounddevice.py
```diff
@@ -2262,13 +2262,68 @@ def reset(self):
...
+
+    Notes
+    -----
+    - The `host_api`, `host_code`, and `host_text` attributes will be None
+      except for the case when `code` == `paUnanticipatedHostError`.
+    
+    Attributes
+    ----------
```
mgeier: What about defining the attributes as class variables as I did in the `default` class?

sounddevice.py
```diff
@@ -2262,13 +2262,68 @@ def reset(self):
...
+    code : int or None
+        The PaErrorCode integer value or None if there is no associated error
+        code.
+    host_api : int or None
+        The index of the host API which returned the error code.
```
mgeier: In the code below, `self.host_api` is actually the host API name.

sounddevice.py
```diff
@@ -2262,13 +2262,68 @@ def reset(self):
...
+        zero-length string.
+    """
+    def __init__(self, msg=None, code=None):
+        self.code = code
+        self.host_code = self.host_api = self.host_text = None
+
+        if code:
+            if code == _lib.paUnanticipatedHostError:
+                info = _lib.Pa_GetLastHostErrorInfo()
+                hostapi = _lib.Pa_HostApiTypeIdToHostApiIndex(info.hostApiType)
```
mgeier: `self.hostapi`?

sounddevice.py
```diff
@@ -2262,13 +2262,68 @@ def reset(self):
...
+                apiinfo = _lib.Pa_GetHostApiInfo(hostapi)
+                self.host_code = info.errorCode
+                if apiinfo: self.host_api = _ffi.string(apiinfo.name).decode()
+                else:       self.host_api = 'unknown'
+                self.host_text = _ffi.string(info.errorText).decode()
+                errormsg = "Unanticipated '{0}' host error"\
+                  .format(self.host_api, self.host_code)
```
mgeier: There's something wrong here ... two variables for one replacement ...?

sounddevice.py
```diff
@@ -2262,13 +2262,68 @@ def reset(self):
...
+    code : int or None
+        The PaErrorCode integer value or None if there is no associated error
+        code.
+    host_api : int or None
+        The index of the host API which returned the error code.
+    host_code : int or None
+        The host error code returned.
+    host_text : str or None
+        A textual description of the host error if available, otherwise a
+        zero-length string.
```
mgeier: What about calling those `hosterror_code` and `hosterror_text`?

What about combining them into a `hosterror_info` tuple, which could be `None` by default?

sounddevice.py
```diff
@@ -2262,13 +2262,68 @@ def reset(self):
...
+    """
+    def __init__(self, msg=None, code=None):
+        self.code = code
+        self.host_code = self.host_api = self.host_text = None
+
+        if code:
+            if code == _lib.paUnanticipatedHostError:
+                info = _lib.Pa_GetLastHostErrorInfo()
+                hostapi = _lib.Pa_HostApiTypeIdToHostApiIndex(info.hostApiType)
+                apiinfo = _lib.Pa_GetHostApiInfo(hostapi)
```
mgeier: What about using `query_hostapis()` here?

sounddevice.py
```diff
@@ -2262,13 +2262,68 @@ def reset(self):
...
+                else:       self.host_api = 'unknown'
+                self.host_text = _ffi.string(info.errorText).decode()
+                errormsg = "Unanticipated '{0}' host error"\
+                  .format(self.host_api, self.host_code)
+            else:
+                errormsg = _ffi.string(_lib.Pa_GetErrorText(code)).decode()
+            errormsg = "{0} [PaErrorCode {1}]".format(errormsg, self.code)
+            if self.host_code is not None:
+                errormsg += ": '{0}' [HostErrorCode {1}]"\
+                  .format(self.host_text, self.host_code)
```
mgeier: What about using the host API name here instead of "HostErrorCode"?
Then we wouldn't need to mention the name above ...

Something like `Unanticipated host error [PaErrorCode -9999]: 'Something went wrong here' [ALSA error 795]`?

mgeier: Or something like `Unanticipated host error [PaErrorCode -9999, ALSA error 795]: 'Something went wrong here'`?

tgarc: Yeah, I like that second suggestion actually.

sounddevice.py
tgarc: I think it's simple enough that it makes sense to document. Would you prefer to discourage people from raising it?

sounddevice.py
tgarc: I think those names are a little clearer. I suppose putting the code and text attributes into a tuple makes things a bit cleaner as well.

tgarc:
Thanks for your review. I think I've addressed all the issues you raised now. I like how this is looking.

sounddevice.py
mgeier: I would probably not explicitly discourage its use, but I simply cannot imagine a situation where a "normal" user would like to raise that exception. Can you think of one?

If somebody wants to implement yet unavailable PortAudio calls, yeah why not, they can use `PortAudioError`, but they sure don't need the docstring to tell them how to instantiate this exception.

sounddevice.py
```diff
@@ -2262,13 +2262,67 @@ def reset(self):
...
+        if code:
+            errormsg = _ffi.string(_lib.Pa_GetErrorText(code)).decode()
+            if code == _lib.paUnanticipatedHostError:
+                info = _lib.Pa_GetLastHostErrorInfo()
+                self.host_api = _lib.Pa_HostApiTypeIdToHostApiIndex(info.hostApiType)
+                hosterror_text = _ffi.string(info.errorText).decode()
+                hosterror_code = info.errorCode
+                self.hosterror_info = hosterror_code, hosterror_text
+                hostname = query_hostapis(self.host_api)['name']
+                errormsg += " [PaErrorCode {0}, {1}ErrorCode {2}]"\
```
mgeier: Thanks for changing this, but there is no such thing as a `XyzErrorCode`, which your spelling would suggest, or is there?
Can't you just use `XYZ error 99` or `XYZ error code 99`, or really just anything without CamelCase?

Just to be clear: `PaErrorCode` is fine, because this actually exists in the PortAudio API.

sounddevice.py
mgeier: I failed to mention this above, but I think if you combine something, it should be the three things, hostapi/hosterror_code/hosterror_text, because they belong together.
One without the other two doesn't make sense, right?

sounddevice.py
```diff
@@ -2262,13 +2262,67 @@ def reset(self):
...
+    hosterror_info = None
+    """Either a tuple containing the host error code and a textual description
+    of the host error or None if the error is not a host API error. The host
+    error description will be a zero length string if one is not available.
+
+    """
+
+    def __init__(self, msg=None, code=None):
+        self.code = code
+        self.hosterror_info = self.host_api = None
```
mgeier: You don't need to set them to `None` because the class variables are already there.

mgeier:
Thanks for bearing with me! Do you happen to know a way to provoke an "unanticipated host error"?
I would like to see how this looks in the wild, and I don't remember ever seeing one.

mgeier:
I found this interesting page about host-API-specific errors: http://www.portaudio.com/docs/proposals/009-HandlingHostAPISpecificErrorCodes.html.

sounddevice.py
```diff
@@ -2262,13 +2262,67 @@ def reset(self):
...
+
+        if code:
+            errormsg = _ffi.string(_lib.Pa_GetErrorText(code)).decode()
+            if code == _lib.paUnanticipatedHostError:
```
mgeier: I think there should still be the substring "unanticipated host error" (or similar) somewhere in the final error message.
Or did you remove this on purpose?

tgarc: The message 'unanticipated host error' is given by GetErrorText on an unanticipatedHostError. The only reason we use to print the message ourselves was to add the name of the host api into the error message. But now we're adding host api details a different way so we can just use the default portaudio error message.

sounddevice.py
tgarc: No there's no such thing as a XyzErrorCode, I was just trying to save characters. I'll come up with an alternative.

tgarc:
| Do you happen to know a way to provoke an "unanticipated host error"?

Yes, but only with my particular set up. I have a loopback device setup with ALSA (snd-aloop) and if I try to open a stream using that device with a different blocksize than is specified in my .asoundrc than alsa will return an error. So here's the example:

`PortAudioError: Error opening Stream: Unanticipated host error [PaErrorCode -9999, AlsaErrorCode -22]: 'Invalid argument'`

sounddevice.py
tgarc: yeah I kind of figured that's what you meant. I agree it makes more sense.

sounddevice.py
tgarc: What if somebody was just wrapping one of the sounddevice functions or classes and wanted to catch a certain exception and re-raise it with their own message? How would they know that the error code argument is to come second? Anyway, that's the use case I had imagined. I'm not sure that anybody would actually need that functionality.

On a related note, reraising a PortAudioError that is a host api error has some unintended repercussions. It would cause GetLastHostError to get called again and I really have no idea if there is even well defined behavior if this happens and much less so if it is the behavior we would want. I had initially left the host error formatting and the call to GetLastHostError inside of `_check` for this very reason.

tgarc:
Yeah I think I had the seen the proposal at some point. Some similar issues were discussed there. The below statement in particular is something that I was concerned about in calling GetLastHostError inside of the PortAudioError constructor:

>A suggestion was made to extend Pa_GetErrorText() so that it retrieved host API specific error strings when a host error occurs, however this was deemed too complicated to implement, especially when multiple host APIs were in use simultaneously.

Technically, the portaudio API is not thread safe but you can have multiple streams running from different audio apis simultaneously. How GetLastHostError would work in the case that multiple host apis were reporting an error is a mystery to me.

sounddevice.py
mgeier: OK, good, thanks!

sounddevice.py
mgeier: OK, I wasn't thinking about that. I don't know if anyone will use this, but I guess there is no reason for not making that possible. There are no disadvantages from our side, right?

But that would mean you'd have to move the initialization code out of the constructor again ...

I guess we could change the constructor to:

```python
__init__(self, msg='', code=None, hosterror_info=None)
```

... and just do all the PortAudio calls outside of it in `_check()`.

Then it should be easy to re-raise with a new message:

```python
try:
    # something
except sd.PortAudioError as e:
    raise sd.PortAudioError('Ayayayay!', e.code, e.hosterror_info) from e
```

Were you thinking about something like that?

mgeier:
> How GetLastHostError would work in the case that multiple host apis were reporting an error is a mystery to me.

I don't know. I hope I never have to use two host APIs at the same time, much less that they raise an error at the same time ...

Anyway, I think this shouldn't be our concern. If we move the error-related calls back to `_check()` (which is always called immediately after the potentially failing API call), we should be fine, right?

tgarc:
> If we move the error-related calls back to _check() (which is always called immediately after the potentially failing API call), we should be fine, right?

I believe that's the case as long as assume single threaded operation ;)

sounddevice.py
tgarc: That was the thought yes

tgarc:
Okay I've pushed what I think are the final commits *but* I ran into an error when I tried to generate the documentation

```
/home/tdos/dev/python-sounddevice/doc/index.rst:19: WARNING: autodoc: failed to import module 'sounddevice'; the following exception was raised:
  File "/home/tdos/dev/python-sounddevice/venv3/lib/python3.4/site-packages/sphinx/ext/autodoc.py", line 551, in import_object
    __import__(self.modname)
  File "/home/tdos/dev/python-sounddevice/sounddevice.py", line 2843, in <module>
    _initialize()
  File "/home/tdos/dev/python-sounddevice/sounddevice.py", line 2791, in _initialize
    _check(_lib.Pa_Initialize(), 'Error initializing PortAudio')
  File "/home/tdos/dev/python-sounddevice/sounddevice.py", line 2709, in _check
    if err == _lib.paUnanticipatedHostError:
AttributeError: 'FakeLibrary' object has no attribute 'paUnanticipatedHostError'
```

I'm not sure how you'd like to handle this one.

sounddevice.py
```diff
@@ -2276,9 +2276,10 @@ class PortAudioError(Exception):
     msg : str, optional
         User error message.
     code : int, optional
-        PaErrorCode value or None if there is no PaErrorCode associated
-        with the error.
-
+        PaErrorCode value.
+    hosterror_info : tuple, optional
```
tgarc: Note that I expect the tuple described here and not the object returned by GetLastHostErrorInfo

sounddevice.py
```diff
@@ -2289,14 +2289,60 @@ def reset(self):
...
-    """This exception will be raised on PortAudio errors."""
+    """This exception will be raised on PortAudio errors.
+
+    Parameters
+    ----------
+    msg : str, optional
+        User error message.
+    code : int, optional
+        PaErrorCode value.
+    hosterror_info : tuple, optional
```
tgarc: @mgeier Note that I expect the tuple described here and not the object returned by GetLastHostErrorInfo

tgarc:
rebased and squashed into single commit

mgeier:
Thanks for the changes!

> I'm not sure how you'd like to handle this one.

I think you can simply add something like this to the `FakeLibrary` in `doc/fake_cffi.py`:

    paUnanticipatedHostError = NotImplemented

> Possibly in another commit?

I don't know. As you prefer.

mgeier:
I think we still have a problem ...

In the "re-raising" scenario, shouldn't it be possible to take the current error message and prepend/append some additional string?

Here's an example, to check if I misunderstood something:

    e1 = sd.PortAudioError('Error playing whatever.wav', -10000)

Raising this exception would generate this message:

    sounddevice.PortAudioError: Error playing whatever.wav: PortAudio not initialized [PaErrorCode -10000]

Now trying to re-raise (is it meant to be used like this?):

    e2 = sd.PortAudioError('MegaPlayer failure: ' + str(e1), e1.code, e1.hosterror_info)

Which would generate this awkward message:

    sounddevice.PortAudioError: MegaPlayer failure: Error playing whatever.wav: PortAudio not initialized [PaErrorCode -10000]: PortAudio not initialized [PaErrorCode -10000]

tgarc:
You're right about the fact that it's not totally straightforward to append
text to the original message. That's partly due to having PortAudioError
format the error string. The other part is that there's no way to grab the
'original' message. You're forced to do something like

`raise PortAudioError('my message: %s' % str(err).split(':')[0], err.code,
err.hosterror_info)`

The simplest solution may be to move all the formatting back into _check.

The other solution would be to use the *args* Exception attribute so that
users would be able to grab the original message. That's how I had it
originally and I think it provides the most flexibility but as you know,
that requires implementing __str__.

At this point, I think I'd lean towards the simpler option. Agreed?

BTW thanks for your patience, I know this has taken much longer than either
of us would have liked.

sounddevice.py
tgarc: I tried documenting the attributes like you said but it doesn't show up in the docs. What am I missing?

sounddevice.py
mgeier: OK, sounds good.

sounddevice.py
mgeier: I did some awkward manipulations in `doc/index.rst`, you'd have to add the `:members:` option:

```rst
.. autoclass:: PortAudioError
   :members:
```

However, the situation has changed since we were discussing this. I'm not sure anymore if it makes more sense to document the class variables instead of listing the attributes in the main docstring as you had done initially. What do you think?

mgeier:
> BTW thanks for your patience

Not at all. Thank *you*. Most pull requesters would've given up already ...
But mind you, we are not done yet!

> The simplest solution may be to move all the formatting back into _check.

Yeah, I think that makes sense.
Sorry that I suggested to move it in the first place.

Using `PortAudioError.args` for more than the error string would probably have made sense when generating the error message inside of the class. Instead of the constructor, it could've been created in the `__str__()` method.
In fact, now that I'm thinking about it, this would still be possible!
We should definitely do the context-dependent PortAudio calls in `_check()`, but the final formatting (including obtaining the host API name if needed) could be done in the `PortAudioError` class. That would actually be a nice separation of concerns, wouldn't it?
We could store something like that in `args`:

```python
args = 'Error opening stream: Internal error', -9986, None
```

```python
args = 'Error opening stream: Unanticipated host error', -9999, (0, -22, 'Invalid argument')
```

... and the complete error string would only be generated in `__str__()`.
In this case we wouldn't need `__init__()` at all, which would make it acceptable again!

We should probably not pad with `None`, but just allow different lengths for `args`:

```python
args = 'Error opening stream: Internal error', -9986
```

```python
args = "You dumb idiot, why didn't you initialize?",
```

Incidentally, this looks very close to what I mentioned in my very first comment above: https://github.com/spatialaudio/python-sounddevice/pull/82#issuecomment-291880038, but the semantics are significantly different.

My re-raising example from above would become:

```python
e2 = sd.PortAudioError('MegaPlayer failure: ' + e1.args[0], *e1.args[1:])
```

I think this situation is slightly more complicated than before but also more flexible. Namely, a user can easily get to the error message without the baked-in numbers.

Having said all that, I'm also completely fine with keeping `PortAudioError.code` and `PortAudioError.hosterror_info`.
But we should decide for either one or the other way.

tgarc:
> We should definitely do the context-dependent PortAudio calls in _check(), but the final formatting (including obtaining the host API name if needed) could be done in the PortAudioError class. That would actually be a nice separation of concerns, wouldn't it?

Yes, I like that idea! I've implemented it and it seems intuitive and more flexible. Having the attribute accessors (.code .hosterror_info) is convenient but since this is kind of advanced usage I think we can get away without having them.

This changes the documentation a bit however. Do you know what's the proper way of documenting a function/constructor with variable length input? This isn't really covered in the numpy documentation guide. Perhaps I should just document the Parameters section as usual and leave it at that?

mgeier:
Thanks, now this looks very nice and clean!

> Do you know what's the proper way of documenting a function/constructor with variable length input?

I don't.

And I think we shouldn't document the constructor at all, since we didn't implement it.
If you think it is helpful, you can add a link to the already existing documentation: https://docs.python.org/3/library/exceptions.html#BaseException.

I think we should only document the `args` attribute, the constructor arguments directly follow from that (you can mention that in the docs if you think it helps understanding).
One very good thing about that is that we don't even have to come up with public names for the components of `args` and names like `code` and `hosterror_info` can stay implementation details.

sounddevice.py
```diff
@@ -2289,14 +2289,40 @@ def reset(self):
...
+    def __str__(self):
+        errormsg = self.args[0] if len(self.args) > 0 else ''
+
+        if len(self.args) > 2:
+            host_api, hosterror_code, hosterror_text = self.args[2]
+            hostname = query_hostapis(host_api)['name']
+            errormsg += " [PaErrorCode {0}, {1} error {2}]"\
+                        .format(self.args[1], hostname, hosterror_code)
+            if hosterror_text:
+                errormsg += ": '{0}'".format(hosterror_text)
```
mgeier: Wouldn't it be actually helpful to show the empty string if there is no host error text?
I don't know if this even can happen, but I think it would be interesting for the user to know that there was an empty error message, wouldn't it?

mgeier: And wouldn't it be better to read if the error number(s) were in the very end?

    PortAudioError: Error opening Stream: Unanticipated host error: 'Invalid argument' [PaErrorCode -9999, ALSA error -22]

tgarc: In light of your 2 comments, I would actually like to have the formatting like this: 
```
PortAudioError: Error opening Stream: Unanticipated host error [PaErrorCode -9999]: 'Invalid argument' [ALSA error -22]
```

This is the most direct mapping of `args` to a formatted string, and will show host_error_message even if it is zero length. It's more consistent than glopping the two error codes into one bracket conditionally. Maybe it's marginally harder to read than having a single listing of error codes at the end of the string, but when you look at it in terms of its components: 

```
pa_error_message [PaErrorCode <xxx>]: host_error_message [<host> error <xxx>]
```

This seems like a more intuitive way of presenting the errors.

mgeier: I agree, this looks good and more consistent.

As an additional plus, this should also even more simplify the implementation.

tgarc:
@mgeier Take a look at how I documented the `args` attribute and let me know if there's anything you'd like changed. Otherwise, I'm done making changes.

mgeier:
Thanks!

I think it's better not to expose names unnecessarily.
Couldn't you document it without mentioning `error_message`, `pa_error_code` and `hosterror_info`?

sounddevice.py
```diff
@@ -2289,14 +2289,41 @@ def reset(self):
...
+
 if not hasattr(_ffi, 'I_AM_FAKE'):
     # This object shadows the 'default' class, except when building the docs.
     default = default()
 
 
 class PortAudioError(Exception):
-    """This exception will be raised on PortAudio errors."""
+    """This exception will be raised on PortAudio errors.
+
```
mgeier: W293 blank line contains whitespace

sounddevice.py
```diff
@@ -2289,14 +2289,41 @@ def reset(self):
...
+        error_message
+            The user error message.
+        pa_error_code
+            The PortAudio `PaErrorCode` value.
+        hosterror_info
+            A 3-tuple containing the host API index, host error code, and
+            the host error message or a zero length string if no message is
+            available.
 
+    """
```
mgeier: There should be an empty line before a method definition. (PEP8?)

sounddevice.py
```diff
@@ -2289,14 +2289,41 @@ def reset(self):
...
+    def __str__(self):
+        errormsg = self.args[0] if len(self.args) else ''
```
mgeier: `len()` is redundant here.

sounddevice.py
```diff
@@ -2289,14 +2289,41 @@ def reset(self):
...
+        if len(self.args) > 1:
+            errormsg = "{} [PaErrorCode {}]".format(errormsg, self.args[1])
```
mgeier: Please use `{0}` and so on for Python 2.6 compatibility.

sounddevice.py
```diff
@@ -2289,14 +2289,41 @@ def reset(self):
...
+        if len(self.args) > 2:
+            host_api, hosterror_code, hosterror_text = self.args[2]
+            hostname = query_hostapis(host_api)['name']
+            errormsg = "{}: '{}' [{} error {}]".format(errormsg, hosterror_text,
```
mgeier: Same here.
You could change the second one from `'{1}'` to `{1!r}`.

sounddevice.py
```diff
@@ -2776,20 +2803,27 @@ def _split(value):
     return invalue, outvalue
 
 
-def _check(err, msg=""):
+def _check(err, msg=None):
```
mgeier: What is the benefit of changing the default from the empty string to `None`?

sounddevice.py
```diff
@@ -2776,20 +2803,27 @@ def _split(value):
...
-        raise PortAudioError(msg)
-    return err
-
+    if err >= 0:
+        return err
+
+    errormsg = _ffi.string(_lib.Pa_GetErrorText(err)).decode()
+    if msg is not None:
+        errormsg = "{0}: {1}".format(msg, errormsg)
+
```
mgeier: W293 blank line contains whitespace

sounddevice.py
```diff
@@ -2776,20 +2803,27 @@ def _split(value):
...
+    if err >= 0:
+        return err
+
+    errormsg = _ffi.string(_lib.Pa_GetErrorText(err)).decode()
+    if msg is not None:
+        errormsg = "{0}: {1}".format(msg, errormsg)
+        
+    if err == _lib.paUnanticipatedHostError:
+        # (gh82) We grab the host error info here since this function should
+        # only be called after a failing API function. This way we can avoid any
```
mgeier: E501 line too long (80 > 79 characters)

sounddevice.py
```diff
@@ -2776,20 +2803,27 @@ def _split(value):
...
-
+    if err >= 0:
+        return err
+
+    errormsg = _ffi.string(_lib.Pa_GetErrorText(err)).decode()
+    if msg is not None:
+        errormsg = "{0}: {1}".format(msg, errormsg)
+        
+    if err == _lib.paUnanticipatedHostError:
+        # (gh82) We grab the host error info here since this function should
```
mgeier: I think a reader who didn't follow our discussion wouldn't know what the alternative to "here" would even be?

sounddevice.py
```diff
@@ -2776,20 +2803,27 @@ def _split(value):
...
+        # only be called after a failing API function. This way we can avoid any
+        # potential issues with multiple simultaneous API usage
+        info = _lib.Pa_GetLastHostErrorInfo()
+        host_api = _lib.Pa_HostApiTypeIdToHostApiIndex(info.hostApiType)
+        hosterror_text = _ffi.string(info.errorText).decode()
+        hosterror_info = host_api, info.errorCode, hosterror_text
+        raise PortAudioError(errormsg, err, hosterror_info)
+
+    raise PortAudioError(errormsg, err)
+
```
mgeier: W293 blank line contains whitespace

sounddevice.py
```diff
@@ -2289,14 +2289,41 @@ def reset(self):
...
+    Attributes
+    ----------
+    args
+        A variable length tuple containing the following parameters when
+        available:
+
+        error_message
+            The user error message.
+        pa_error_code
+            The PortAudio `PaErrorCode` value.
```
mgeier: ` ``PaErrorCode`` `

doc/fake_cffi.py
```diff
@@ -18,6 +18,7 @@ class FakeLibrary(object):
     # from portaudio.h:
 
     paFloat32 = paInt32 = paInt24 = paInt16 = paInt8 = paUInt8 = NotImplemented
+    paUnanticipatedHostError = NotImplemented
```
mgeier: This is not necessary anymore, right?

tgarc: correct !

sounddevice.py
tgarc: That's why I put the github issue number, but I can reword it a bit.

sounddevice.py
tgarc: if default msg="" and no message is given you get: 

```PortAudioError: : Invalid number of channels [PaErrorCode xxx]```

With None, that extra semicolon + space doesn't get printed. I could also ignore it if bool(msg) evaluates to False, but I prefer not to assume anything about msg except that if it is None it should be ignored. In this particular case it is unlikely ever to be an issue, but I consider it good practice.

File: sounddevice.py (hunk 28 above)
Comment (tgarc) [inline sounddevice.py]: Didn't know about that. Thanks

File: sounddevice.py (hunk 29 above)
Comment (tgarc) [inline sounddevice.py]: Again this is a coding habit. '{1}' is more explicit than assuming {1!r} will add quotes around the object. And in fact with Python2 this will cause a `u` (unicode string) to be prepended to the error message which I don't care for (but maybe this is what you intended?).

**Comment by tgarc** [comment]
> Sorry for being a pain in the ass, but I found a few more nits to pick ...

hah no worries. I want this to be up to snuff.

>  Couldn't you document it without mentioning error_message, pa_error_code and hosterror_info?

I was having a little trouble wording it without giving them a label but I'll see what I can do.

File: sounddevice.py (hunk 29 above)
Comment (mgeier) [inline sounddevice.py]: I guess both are OK and both have their flaws.

I wasn't aware of the `u` prefix, but my policy regarding Python 2 is: I want it to work (for now, until support is dropped), but I don't care if it is ugly.

For me, using `{1!r}` is habit.
The reason for that is that the quotes automatically change if the strings contains single quotes.
The downside is that "special" characters like `\n` are shown with their escape character ... which is good in many cases, but here it's probably not ...

So in the end it looks like your way is better here!

File: sounddevice.py (hunk 30 above)
Comment (mgeier) [inline sounddevice.py]: I also don't think it'll be an issue here, for me it's also about good practice.

In your version, I get exactly the awkward output you mentioned if I use `_check(xxx, '')` (which of course I wouldn't do in practice, but still).

I think it's more reasonable (and more Pythonic, too) to ignore it if it is *falseish*.

This is IMHO very idiomatic:

```python
if msg:
    # do something
```

I would use `None` only if there are valid falseish values which I care about (and if the default value is something mutable, which might be problematic in some cases).

sounddevice.py
mgeier: Thanks, the new wording is much clearer!

sounddevice.py
```diff
@@ -2289,13 +2289,40 @@ def reset(self):
...
-    """This exception will be raised on PortAudio errors."""
+    """This exception will be raised on PortAudio errors.
+
+    Attributes
+    ----------
+    args
+        A variable length tuple containing the following elements when
+        available:
+
+        1) The user error message
```
mgeier: What does "user" mean here?

sounddevice.py
```diff
@@ -2289,13 +2289,40 @@ def reset(self):
...
+        2) The PortAudio ``PaErrorCode`` value
+        3) A 3-tuple containing the host API index, host error code, and the
+           host error message or a zero length string if no message is
+           available
```
mgeier: The thing about the zero length string seems a bit verbose.
Is it really relevant to mention it?
Does this even happen?

If yes, wouldn't something like this be easier to read?

    ... and the host error message (which may be an empty string)

A combination of "and" and "or" is often hard to understand in natural language.
In fact "and" binds stronger than "or", which makes the sentence actually wrong!

mgeier:
BTW, I just discovered a nice Python feature: If `__str__()` throws an exception, this is not shown in the traceback and the exception doesn't replace the original one.
Instead, we simply get this:

    sounddevice.PortAudioError: <exception str() failed>

I was worried what happens if `__str__()` throws (e.g. if `query_hostapis()` doesn't work), but as so often, Python just does the right thing!

sounddevice.py
tgarc: Yes, I agree that's a bit clearer.

sounddevice.py
tgarc: heh, actually it's kind of vague. I meant "the error message that get's presented to the user" but that's not really correct either. I didn't want to say "The PortAudio error message" since that's also not totally correct, but maybe it's close enough. Would that be alright?

mgeier: Well as we currently use it, this is most of the times a combination of a message given by the `sounddevice` module and a message given by PortAudio. Sometimes it's only the former, sometimes it's only the latter.

I think we can stay vague and just write "The error message" or "A string describing the error" or something like that.

tgarc:
I've committed something that addresses all the issues raised.

mgeier:
Thanks a lot @tgarc! And sorry for being so nit-picky ...

tgarc:
hooray! That's alright. I appreciate your attention to detail.

PR #99, opened 2017-08-19
close() sets the internal stream pointer to NULL

mgeier:
I split this into three commits, because they have quite different implications.

I think setting the stream pointer to NULL (abf1f19839059453436738d8d2979d22772c101c) has no negative implications.
AFAICT, the error messages are the same as when using an invalid pointer after `close()`. But it might be safer on some systems (see mailing list link below).

The `.closed` property (8370db2cf4ff08a23119824f2cecd08ad15a1165) seems harmless, Python file objects tend to have such a thing, so it shouldn't be too uncommon.

The last commit (9af4e43a387e312e1c79e35eacb65b138b816c83) actually changes the behavior. Instead of throwing an exception, it now returns `active == False` on closed streams (without actually checking activeness).
I think that is logically justifiable, because a closed stream can never ever be active, but I'm not quite sure.

The background to all this is that a derived class might have some state (as "private" instance variables) that's written by the callback function (possibly in C and without holding the GIL!). While the callback is running, this shouldn't be accessed, but afterwards it's OK. Even after the stream was closed.
If I'm not mistaken, this cannot be checked with the current public API, but it could be checked with a combination of `.closed` and `.active`. Or, with the third commit, this could simply be checked by `.active` alone.

@tgarc You seem to be an expert in that area, what do you think about this?

See also https://lists.columbia.edu/pipermail/portaudio/2017-March/001074.html.

### DIFF
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -1181,6 +1181,11 @@ class _StreamBase(object):
         return _check(_lib.Pa_IsStreamStopped(self._ptr)) == 1
 
     @property
+    def closed(self):
+        """``True`` after a call to `close()`, ``False`` otherwise."""
+        return self._ptr == _ffi.NULL
+
+    @property
     def time(self):
         """The current stream time in seconds.
 
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -1162,6 +1162,8 @@ class _StreamBase(object):
         stopped
 
         """
+        if self.closed:
+            return False
         return _check(_lib.Pa_IsStreamActive(self._ptr)) == 1
 
     @property
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -1286,6 +1286,7 @@ class _StreamBase(object):
 
         """
         err = _lib.Pa_CloseStream(self._ptr)
+        self._ptr = _ffi.NULL
         if not ignore_errors:
             _check(err, 'Error closing stream')
 
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -1180,6 +1180,8 @@ class _StreamBase(object):
         active
 
         """
+        if self.closed:
+            return True
         return _check(_lib.Pa_IsStreamStopped(self._ptr)) == 1
 
     @property


tgarc:
Hey mgeier. The `closed` property seems like a nice addition. What was the motivation to add these changes though? These behaviors kind of diverge from the PortAudio stream's behavior.

I was interested in making changes like this because I wanted to use Streams in a multi-threaded context where I wanted to do things like for example abort the stream in the main thread if it had not already been closed by the subthread. (Of course, I still had to add a stream mutex to make these functions thread safe.)

Either way, I do think the `closed` property is a useful feature. And I can't see anything about these changes that could be harmful or limiting to users. I'm in favor of these changes 👍

mgeier:
Thanks @tgarc, I very much appreciate your feedback!

I tried to hint on the general motivation above, but I'll make it more concrete: in [rtmixer](https://github.com/spatialaudio/python-rtmixer) (currently in the `fetch-stats` branch) I'm planning to add a `.stats` attribute (a property, to be more exact) to get e.g. the number of xruns that happened during the runtime of the callback. This attribute should not be queried while the callback is running, because the underlying structure is not protected by a lock or anything. For this reason, I'm checking `stream.active` in the implementation of the property.

So far so good. But I would like to use it somewhat like that:

```python
with rtmixer.Mixer() as stream:
    # run the stream for some time
print(stream.stats.output_underflows)
```

The problem is that exiting the context manager calls `close()` (which I think makes sense), which invalidates the stream pointer, and the check for `stream.active` that's done while getting `stream.stats` will raise an exception.

> These behaviors kind of diverge from the PortAudio stream's behavior.

Yeah, that's why I'm a bit unsure about this.

It is a change of behavior, but this change is either invisible to users (setting the pointer to NULL) or only changes errors into non-errors.

mgeier:
PS: I would like to add that the checking `.closed` in `.active` is really only a convenience, if we don't do this, users would just have to do this on their own, which wouldn't be a big problem, but I think it would be a bit annoying.

I also want to point out that `.active` and `.stopped` would be somewhat asymmetrical in my current proposal, because `.stopped` would still raise an exception after `.close()`.
Initially, I thought that after `.close()` we cannot tell if a stream was "stopped". But now that I'm thinking about it, there might also be a logic argument for that: the documentation says that `.close()` behaves as if `abort()` had been called, therefore a `.closed` stream must always be `.stopped`, right?

Shall I adapt the `.stopped` property, too?

tgarc:
I would adapt the `.stopped` property like you said to be more consistent.

In general I would say that if it ain't broke don't fix it, but I think the change in behavior is okay and will make things a little more robust.

PR #103, opened 2017-10-15
Move sounddevice.py to src/

mgeier:


tgarc:
I'm pretty indifferent to these kinds of aesthetic changes but adding a src folder feels counter intuitive. Python packages pretty much always have the main source file in the respository, or in a subfolder that has the same name as the package. A src/ folder reminds me more of C projects.

mgeier:
Some time ago I would have had a similar opinion, but then I was enlightened by https://blog.ionelmc.ro/2014/05/25/python-packaging/ which was recommended by https://docs.pytest.org/en/latest/goodpractices.html.
I'm not repeating the arguments here, but I think they make a pretty good case for using `src/`.

And it would definitely not be a purely aesthetic change.

Of course there are also other opinions, e.g. http://as.ynchrono.us/2007/12/filesystem-structure-of-python-project_21.html, which says:

> Don't [...] put your source in a directory called `src` or `lib`. This makes it hard to run without installing.

* This is not quite true (see the comments: http://as.ynchrono.us/2007/12/filesystem-structure-of-python-project_21.html?showComment=1324403776492&m=1#c8055206121684953974
... where it turns out that it is actually quite easy and the argument reduces to "it's still pointless extra work")
* It is the point of the abovementioned good practices that we *shouldn't* run stuff without installing
* When using CFFI out-of-line mode, this is a moot point, because we *have to* install the module anyway.

And the page http://docs.python-guide.org/en/latest/writing/structure/ simply says:

> Your library does not belong in an ambiguous src or python subdirectory.

... which isn't an argument at all.

So I'm definitely leaning towards moving to `src/`, but I might have missed something. Are there other arguments?

tgarc:
hah, thorough as always @mgeier . I can see now where using a src directory would be helpful, even just to help avoid mistakes that happen when the python module is at the root directory. I will probably start doing this myself. Thanks for the lesson!

PR #104, opened 2017-10-17
Added support for unicode strings on python 2.x

roel0:
Added support for:

```
import sounddevice as sd
sd.query_devices("Scarlet")
sd.query_devices(u"Scarlet")
```

### DIFF
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -2497,7 +2497,7 @@ def _array(buffer, channels, dtype):
 
 def _split(value):
     """Split input/output value into two values."""
-    if isinstance(value, str):
+    if isinstance(value, basestring):
         # iterable, but not meant for splitting
         return value, value
     try:
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -60,6 +60,11 @@ _sampleformats = {
 _initialized = 0
 _last_callback = None
 
+try:
+  basestring
+except NameError:
+  basestring = str
+
 
 def play(data, samplerate=None, mapping=None, blocking=False, loop=False,
          **kwargs):
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -61,9 +61,9 @@ _initialized = 0
 _last_callback = None
 
 try:
-  basestring
+    _basestring = basestring
 except NameError:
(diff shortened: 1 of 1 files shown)


sounddevice.py
```diff
@@ -60,6 +60,11 @@
 _initialized = 0
 _last_callback = None
 
+try:
+  basestring
+except NameError:
+  basestring = str
```
tgarc: Actually basestring in python 3 is a combination of (bytes, str)

mgeier: We could do it like in the [soundfile](https://github.com/bastibe/PySoundFile/blob/cc7c0dd786c23a1aa88b64ce25d901bfaf514b38/soundfile.py#L18-L21) module:

```python
try:
    _unicode = unicode
except NameError:
    _unicode = str
```

and then

```python
if isinstance(value, (_unicode, bytes)):
    ...
```

mgeier: On second thought, some of the logic might not work on Python 3 `bytes`, right?
And it doesn't have to ...

tgarc: I would tend to do 

``` python
try:
    _basestring = basestring
except NameError:
    _basestring = (str, bytes)
```

This works fine if your just doing instance checks

roel0: Or a one line solution to support string, unicode and bytes on both python2 and python3:

     isinstance(value, ("".__class__, u"".__class__, b"".__class__))

mgeier: @roel0 The "u" prefix isn't available in Python 3.0-3.2 (where I think it causes a `SyntaxError`), but that's probably not a problem.

Regardless, I like @tgarc's suggestion (sans parentheses), since `basestring` is just meant for `isinstance()` checks anyway.
Note, however, that `query_devices()` doesn't work with `bytes` in Python 3. And I think that's OK. It really shouldn't work with `bytes`.
Still, I think it's OK to allow `bytes` in `_split()`.

roel0: Replaced with @tgarc's suggestion

mgeier:
Thanks @roel0 and @tgarc!

Issue #166, opened 2019-01-30
sounddevice silently redirects stderr to /dev/null


pytaunay:
This is originally in reference to an issue in the psychopy Sound backend which uses sounddevice (see https://github.com/psychopy/psychopy/issues/2230)

`sounddevice` has a few lines of code that were added to remove the standard error output for convenience (pull request #12 in response to issues #11 and #10):
```
2670 def _ignore_stderr():
2671     """Try to forward PortAudio messages from stderr to /dev/null."""
2672     try:
2673         stdio = _ffi.dlopen(None)
2674         devnull = stdio.fopen(_os.devnull.encode(), b'w')
2675     except (OSError, AttributeError):
2676         return
2677     try:
2678         stdio.stderr = devnull
2679     except _ffi.error:
2680         try:
2681             stdio.__stderrp = devnull
2682         except _ffi.error:
2683             stdio.fclose(devnull)
2684 
2685 
2686 _ignore_stderr()
```
This redirects all of the standard error output to /dev/null and subsequently breaks built-in functions that use stderr display (e.g. "input"). This occurs in non-interactive sessions only (probably because of the way the CPython backend handles display with a prompt in interactive sessions).

mwe.py:
```
#!/usr/bin/python3

# This will display "Enter data" on the terminal
t = input("Enter data\n")

import sounddevice

# This does not 
t = input("Enter data\n")
```
`python3 mwe.py` will display "Enter data" the first time, but not the second time after sounddevice is imported. Commenting line 2686 in sounddevice.py gets rid of the issue.

stderr should **not** be redefined as /dev/null internally. I would suggest removing the `_ignore_stderr` function. For those who do not like stderr output, just use piping externally to redirect stderr to /dev/null or a log file.
Piping example:
```
#!/usr/bin/python3

# This will display "Enter data" on the terminal
t = input("Enter data\n")

print("Test")
```
`python3 pipe.py 2> /dev/null` will display only "Test". "Enter data" gets redirected to /dev/null.

hoechenberger:
Related issue with the same cause: #37

mgeier:
Thanks for the report!

First of all, I'm not planning to remove the "ignore stderr" feature by default, but I'm open to suggestions for making it optional.

This feature is very important, because for PyAudio (which doesn't have this feature), those warning messages cause a lot of confusion, as shown by these SO questions:

[...]

PR #168, opened 2019-02-04
Implement stderr cleanup using context manager

pytaunay:
Addresses issue #166 :

- Redirect warning messages to a temporary log file
- Restore stderr to its default file descriptor so that built-in functions such as "input" and other possible error messages are not impacted
- Open-up the temporary log file to users so that they can also log additional stderr messages

Example:
```
# Import ctypes to expose "write"
import ctypes
libc = ctypes.CDLL(None)

input("Testing! ") # Shows "Testing!"

import sounddevice # Shows where the log file is created and logs the warnings from e.g. ALSA

input("More testing\n") # Now shows "More testing" (did not use to)
print("I am on stdout!") # Stdout is not impacted

# We can also ignore further messages, including directly from libc
with sounddevice.stderr_capture(sounddevice.tf.name) as cap:
    s = "Tests\n"
    libc.write(2,s,4*len(s))
    input("I am suppressed!\n")

input("I am not suppressed!\n") # But once out of the context everything is back to normal
```

Output on Ubuntu 16.04:
```
> python3 ~/foo3.py
Testing! 
Sounddevice: Logging error messages in  /tmp/tmpkfbf7wau.sounddevice
More testing

I am on stdout!

I am not suppressed!
> cat /tmp/tmpkfbf7wau.sounddevice
ALSA lib pcm.c:2266:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.rear
ALSA lib pcm.c:2266:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.center_lfe
ALSA lib pcm.c:2266:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.side
ALSA lib pcm_route.c:867:(find_matching_chmap) Found no matching channel map
ALSA lib pcm_route.c:867:(find_matching_chmap) Found no matching channel map
ALSA lib pcm_route.c:867:(find_matching_chmap) Found no matching channel map
ALSA lib pcm_route.c:867:(find_matching_chmap) Found no matching channel map
Cannot connect to server socket err = No such file or directory
Cannot connect to server request channel
jack server is not running or cannot be started
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for 4294967295, skipping unlock
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for 4294967295, skipping unlock
Tests
I am suppressed!
```

Tested on Ubuntu 16.04 with Python 3.5.2 and Mac OS 10.13.6 (High Sierra) with Python 3.6.2

### DIFF
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -56,9 +56,11 @@ import platform as _platform
 import sys as _sys
 from ctypes.util import find_library as _find_library
 from _sounddevice import ffi as _ffi
-import tempfile as _tf
+import tempfile
 from contextlib import contextmanager
 
+# Create a temporary file and signal where we are logging error messages 
+_tf = tempfile.NamedTemporaryFile(mode="w+",suffix=".sounddevice",delete=False)
 
 try:
     for _libname in (
@@ -2642,9 +2644,10 @@ def _get_device_id(id_or_query_string, kind, raise_on_error=False):
 
 def _initialize():
     """Initialize PortAudio."""
-    global _initialized
-    _check(_lib.Pa_Initialize(), 'Error initializing PortAudio')
-    _initialized += 1
+    with stderr_capture(_tf.name) as cap:
+        global _initialized
+        _check(_lib.Pa_Initialize(), 'Error initializing PortAudio')
+        _initialized += 1
 
 
 def _terminate():
@@ -2713,12 +2716,9 @@ def stderr_capture(to=_os.devnull,stderr=None):
             stderr.flush()
             _os.dup2(copied.fileno(), stderr_fd)  # $ exec 2>& copied
 
-# Create a temporary file and signal where we are logging error messages 
-tf = _tf.NamedTemporaryFile(mode="w+",suffix=".sounddevice",delete=False)
 
-with stderr_capture(tf.name) as cap:
-    _atexit.register(_exit_handler)
-    _initialize()
+_atexit.register(_exit_handler)
+_initialize()
 
 if __name__ == '__main__':
     print(query_devices())
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -2715,7 +2715,6 @@ def stderr_capture(to=_os.devnull,stderr=None):
 
 # Create a temporary file and signal where we are logging error messages 
 tf = _tf.NamedTemporaryFile(mode="w+",suffix=".sounddevice",delete=False)
-print("Sounddevice: Logging error messages in ", tf.name)
 
 with stderr_capture(tf.name) as cap:
     _atexit.register(_exit_handler)
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -56,6 +56,8 @@ import platform as _platform
 import sys as _sys
 from ctypes.util import find_library as _find_library
 from _sounddevice import ffi as _ffi
+import tempfile as _tf
+from contextlib import contextmanager
 
 
 try:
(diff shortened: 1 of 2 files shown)


mgeier:
Thanks for this PR!

sounddevice.py
```diff
@@ -2666,26 +2668,58 @@ def _exit_handler():
...
+            yield stderr # Allow code to be run with the redirected stderr
+        finally:
+            # Restore stderr to its previous value
+            # NOTE: dup2 makes stderr_fd inheritable unconditionally
+            stderr.flush()
+            _os.dup2(copied.fileno(), stderr_fd)  # $ exec 2>& copied
+
+# Create a temporary file and signal where we are logging error messages 
+tf = _tf.NamedTemporaryFile(mode="w+",suffix=".sounddevice",delete=False)
+print("Sounddevice: Logging error messages in ", tf.name)
```
mgeier: This kinda defeats the purpose of wanting to avoid annoying terminal output, doesn't it?

pytaunay: Fair enough. I just put that there to make sure one can grab the error messages if necessary. I believe it's a bit more informative than the stuff that ALSA spits out for example.

The print line can be removed. You can then mention in the documentation that the error stream is logged in the temporary folder and that its location can be accessed with sounddevice.tf

mgeier: Can you please remove the output? Then I can continue reviewing.

mgeier:
Thanks @pytaunay for the update!

Your PR changes two separate things:

* forward to a temporary file instead of `/dev/null`
* apply the forwarding only while PortAudio is initializing

Do you really need both?

I have the feeling that the second thing would suffice.

Also, your solution doesn't work if a user forwards `stderr` to something else (see https://docs.python.org/3/library/contextlib.html#contextlib.redirect_stderr).
E.g. this raised an exception when I tried it:

```python
import contextlib
import io
f = io.StringIO()
with contextlib.redirect_stderr(f):
    import sounddevice
```

This doesn't raise an exception but the redirection doesn't seem to work:

```python
import contextlib
with open('mylogfile.txt', 'w') as f:
    with contextlib.redirect_stderr(f):
        import sounddevice
```

Also, if you are limiting the forwarding the the PortAudio initialization, you should activate it in `_initialize()`. This way it also works if somebody uses `sd._terminate()` and `sd._initialize()` to re-initialize PortAudio (which is actually sometimes needed).

pytaunay:
> Your PR changes two separate things:
> 
> * forward to a temporary file instead of `/dev/null`
> * apply the forwarding only while PortAudio is initializing
> 
> Do you really need both?

I would argue that I would need both. The warning messages are still important to me. One can refer back to the temp file that contains them with this method. 

> Also, your solution doesn't work if a user forwards `stderr` to something else (see https://docs.python.org/3/library/contextlib.html#contextlib.redirect_stderr).
> E.g. this raised an exception when I tried it:
> 
> ```python
> import contextlib
> import io
> f = io.StringIO()
> with contextlib.redirect_stderr(f):
>     import sounddevice
> ```
> 
> This doesn't raise an exception but the redirection doesn't seem to work:
> 
> ```python
> import contextlib
> with open('mylogfile.txt', 'w') as f:
>     with contextlib.redirect_stderr(f):
>         import sounddevice
> ```
contextlib.redirect_stderr does not redirect at the file descriptor level, much like its counterpart redirect_stdout. That's why output called from C libraries just passes through and a context manager is needed. I will look into the reason why the redirection from another file does not work correctly.

> Also, if you are limiting the forwarding the the PortAudio initialization, you should activate it in `_initialize()`. This way it also works if somebody uses `sd._terminate()` and `sd._initialize()` to re-initialize PortAudio (which is actually sometimes needed).

Sounds good.

mgeier:
@pytaunay 

> I will look into the reason why the redirection from another file does not work correctly.

Any news on that?

pytaunay:
I do apologize; the past two months have been quite hectic!

Contextlib uses `setattr` to modify `sys` directly. In the class `_RedirectStream` the `__enter__` function calls `setattr`:
```
185 class _RedirectStream(object):
186 
187     _stream = None
188 
189     def __init__(self, new_target):
190         self._new_target = new_target
191         # We use a list of old targets to make this CM re-entrant
192         self._old_targets = []
193 
194     def __enter__(self):
195         self._old_targets.append(getattr(sys, self._stream))
196         setattr(sys, self._stream, self._new_target)
197         return self._new_target
198 
199     def __exit__(self, exctype, excinst, exctb):
200         setattr(sys, self._stream, self._old_targets.pop())
```
The redirect_stderr class from contextlib just sets `_stream` to "stderr". So what happens when using the following piece of code ` with contextlib.redirect_stderr(f)` is that only sys.stderr gets overwritten. However, as we've seen before, this is not enough to deal with the C backend. When the `stderr_capture` context manager from this PR uses `_sys.stderr` it then only sees the file descriptor of the file "f", not the true standard error file descriptor. Hence why you still see output on the console.

I would simply recommend not using contextlib wrappers since I am not sure how to grab the original sys.stderr from within the new context.

I am also pushing a solution based on your suggestion of putting the stderr_capture only in the initialize function.

mgeier:
Thanks for the update!

I still don't like the idea of creating a temporary file at all.

More so, the temporary file is also created on Windows, where the whole thing isn't necessary at all.

> I would simply recommend not using contextlib wrappers

Well I can't control what users of the `sounddevice` module do before `import`ing the module.
And using the `contextlib` contextmanagers is what's recommended in the Python docs, so it's definitely to be expected that people are using them.

Anyway, I would prefer if it's not so easy to break things.

I've created an alternative attempt in #176, what do you think about that?

pytaunay:
Your solution seems to be working better. I have tried it on the test case I have above and on the context manager suggestion you had. It does redirect to /dev/null correctly and does not break either the context manager or the input() function. 

Since stderr is modified only during initialization it does not seem to affect anything after sounddevice is imported. Looks great!

mgeier:
Thanks, I've merged #176.

Issue #205, opened 2019-10-13
Break down the API Documentation page


cool-RR:
I'm using the API Documentation page and its length is make it difficult to use. Streams are a complicated subject, and it doesn't help that their documentation is on the same page with the other features.

In the beginning of the page you have an "API overview" section with 7 bullets, I suggest that the division to pages be roughly as that of the bullets.

mgeier:
That's a great idea, would you like to make a PR for that?

The main API page could have this (without `:members:`!):

    .. automodule:: sounddevice

And then a `toctree` directive with the sub-pages.

Most of [the current API page](../blob/afe8fcc4134fe2b9a5f89c15c76fc2b67c15a3d4/doc/api.rst) is already split into separate pieces, so it should be easy to create separate pages.

cool-RR:
Can you please give me the code needed to build the documentation locally?

I tried this snippet that I found in `.travis.yml`:

```
python setup.py build_sphinx -b linkcheck -W
```

But I got this error:

```
C:\Users\Administrator\Documents\Python Projects\Third-party\sounddevice\doc\index.rst:5:toctree conta
ins reference to document 'CONTRIBUTING' that doesn't have a title: no link will be generated         
```

mgeier:
The command you used doesn't actually build the HTML files, it only checks if all the links are still valid.

Here are the instructions for building the docs: https://github.com/spatialaudio/python-sounddevice/blob/master/CONTRIBUTING.rst

mgeier:
This was fixed in #207.

PR #207, opened 2019-10-14
Split API documentation to multiple pages (#205)

cool-RR:


cool-RR:
@mgeier Build failure related to my change?

doc/api/index.rst
```diff
@@ -0,0 +1,20 @@
+API Documentation
+=================
+
+.. toctree::
```
mgeier: I think it's better if the module docstring comes first, and afterwards the sub-pages.

doc/api/index.rst
```diff
@@ -0,0 +1,20 @@
...
+   misc
+   settings
+
+.. automodule:: sounddevice
+   :undoc-members:
+   :exclude-members: RawInputStream, RawOutputStream, RawStream,
+                     InputStream, OutputStream, Stream,
+                     CallbackFlags, CallbackStop, CallbackAbort,
+                     PortAudioError, DeviceList,
+                     AsioSettings, CoreAudioSettings, WasapiSettings
```
mgeier: None of those options have an effect, right?

I think they can all be removed.

doc/api/misc.rst
```diff
@@ -0,0 +1,13 @@
+API Documentation: Miscellaneous
```
mgeier: No need to repeat "API Documentation". That should be clear from context.

doc/api/functions.rst
```diff
@@ -0,0 +1,11 @@
+API Documentation: Functions
```
mgeier: I think "Functions" isn't really a meaningful category.

Probably "Convenience Functions using NumPy Arrays" would be a relevant category?

doc/api/settings.rst
```diff
@@ -0,0 +1,8 @@
+API Documentation: Settings
```
mgeier: Those are not any settings, those are "Platform-specific Settings"

mgeier:
> Build failure related to my change?

Probably. Let's see if it goes away after your next changes.

doc/api/index.rst
cool-RR: Done.

doc/api/misc.rst
cool-RR: Done.

doc/api/settings.rst
cool-RR: Done

doc/api/functions.rst
cool-RR: I changed to "Top-level functions".

I think that "convenience functions" is terminology that puts the point of view of the developer higher than the point of view of the user. As the developer, you see the main part of `sounddevice` as the complex stream class hierarchy and the integration with portaudio. But for the average user, `play` and `rec` are going to be the most important things, and all the rest are things under the hood that you can use if you want to really customize how things work.

That's my opinion, if you feel otherwise I'll change to whichever title you'd pick.

cool-RR:
The errors in the CI are indeed related to my changes, at least the first one that I now fixed by adding `:noindex:`, but I can't figure out the second one: 

```
/home/travis/build/spatialaudio/python-sounddevice/sounddevice.py:docstring of sounddevice:4:'any' reference target not found: play()
```

Any idea? I see that `play` is mentioned in the module docstring which is why Sphinx is looking for it, but the `play` function is defined right there, I don't know why it won't find it. I'm not sure whether it's related to my change.

doc/api/functions.rst
mgeier: There have been multiple questions (e.g. on stackoverflow) in the past, where users were using `play()` et al. for situation where they shouldn't be used. So no, I don't agree, this separation/grouping is very much relevant for "users". That's also the reason why I recently did e8253a615d6f70f7a1f6bc662e703a857fd2e8c4, where I tried to point out the limitations of those functions.

I *do* agree that `play()` is the most important function for most people, but I think it is currently sufficiently visible.

Anyway, we can postpone the naming of the page for a moment.

More importantly, the "convenience functions" should be on a page of their own, not intermixed with the "low level" wrapper functions.

doc/api/functions.rst
```diff
@@ -0,0 +1,12 @@
+Top-level functions
+===================
+
+.. automodule:: sounddevice
```
mgeier: Please don't repeat the module docstring in this page.
I think it's sufficient to mention it in `api/index.rst`.

This should also solve one of the warnings.

mgeier:
I think the problem is that `autodoc` doesn't know which module we are talking about in the sub-pages.

I think this could help: http://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-py:currentmodule

doc/api/index.rst
cool-RR: Done.

doc/api/functions.rst
cool-RR: Okay, I changed to the name you suggested. I broke it down into two pages, I'm not sure I got the exact division you meant, feel free to point out anything you want me to move.

doc/api/functions.rst
cool-RR: Done.

cool-RR:
Hmm, I pushed a new version but I'm still getting this:

```
/home/travis/build/spatialaudio/python-sounddevice/sounddevice.py:docstring of sounddevice.play:47:'any' reference target not found: default.samplerate
```

doc/api/index.rst
mgeier: You left one option in there?

mgeier:
You should use the `:members` flag when documenting `default`, otherwise the attributes will not be documented and the links will not work.

doc/api/convenience_functions.rst
```diff
@@ -0,0 +1,14 @@
+Convenience functions
+=====================
+
+.. py:currentmodule:: sounddevice
```
mgeier: You can skip the `py:` as it is the default.

mgeier:
Please make sure to capitalize your titles as in the rest of the docs.

doc/api/convenience_functions.rst
cool-RR: Done.

doc/api/convenience_functions.rst
```diff
@@ -0,0 +1,14 @@
...
+
+.. autofunction:: play
+
+.. autofunction:: rec
+
+.. autofunction:: playrec
+
+.. autofunction:: wait
+
+.. autofunction:: stop
```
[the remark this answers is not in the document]
cool-RR: Done.

doc/api/index.rst
cool-RR: Ah, right. Removed.

cool-RR:
Woohoo, CI is passing!

cool-RR:
Let me know whether anything else is needed to merge this.

mgeier:
`sounddevice.default` doesn't really fit into "low-level functions" because it's a class, right?

I don't really like the name "low-level functions", can you come up with a better one?

I don't know if it would fit, but what about moving `DeviceList` into the same page as `query_devices()` because they are closely related?

BTW, feel free to change the module docstring to match the API sub-pages.

cool-RR:
I think we should push as much as possible to a separate PR after this PR is passed, there's no end to how much we can tweak it.

I don't have a better name than "Low-level functions", and to be honest I don't really understand the division between "Miscellaneous" and "Low-level functions." As far as I'm concerned I'd put both on the same page called "Miscellaneous".

I moved `DeviceList` now.

I tried moving `default` to its own page but it results in Sphinx doing weird linking and formatting, and I couldn't figure out why.

mgeier:
> I think we should push as much as possible to a separate PR after this PR is passed, there's no end to how much we can tweak it.

I don't actually think there is much left to tweak.
It's just about what goes where and what titles to use.
This is the whole idea of this PR, I don't see how a new PR would help with that.

> As far as I'm concerned I'd put both on the same page called "Miscellaneous".

I hope we find a better solution.

> I tried moving `default` to its own page but it results in Sphinx doing weird linking and formatting, and I couldn't figure out why.

Can you please show what went wrong?
We can probably fix the problems.

Did you happen to call the new page `default.rst`?
This would cause name clashes, you could e.g. call it `module-defaults.rst`, then it should work.

mgeier:
What about something like this?

* Convenience Functions using NumPy Arrays
* Checking Available Hardware
* Module-wide Default Settings
* Platform-specific Settings
* Streams using NumPy Arrays
* Raw Streams
* Miscellaneous

cool-RR:
> Did you happen to call the new page `default.rst`?
> This would cause name clashes, you could e.g. call it `module-defaults.rst`, then it should work.

That did the trick :)

> I hope we find a better solution.

How about something with "hardware"? Like "Hardware functions" or "Low-level hardware functions". It'll help if you could explain what about the "Low-level functions" name bothers you, I think it's informative.

cool-RR:
Oops, looks like we sent a message at the same time. I'll change to your new names, thanks.

mgeier:
BTW, I would very much prefer if the file names use hyphens (`-`) instead of underscores (`_`). Just a personal preference ...

cool-RR:
All done and tests pass :)

mgeier:
Two more things:

* I think it would be nicer if `DeviceList` comes after `query_devices()`.
* `sleep()` and `get_portaudio_version()` are not really checking for hardware, I guess it would be better to move them to "Miscellaneous"?

cool-RR:
Done and done.

mgeier:
Perfect, thanks a lot!

cool-RR:
:beer:

cool-RR:
Any idea why this bad boy isn't online yet?

https://python-sounddevice.readthedocs.io/en/0.3.14/api.html

mgeier:
You should check out "latest" (which is a build of the `master` branch): https://python-sounddevice.readthedocs.io/en/latest/api/index.html

cool-RR:
Cool, thanks :)

PR #214, opened 2019-12-02
DOC: Add autosummary to each API sub-page

mgeier:
This is an "add-on" to #207.

@cool-RR What do you think about this?


cool-RR:
I compared one page ("Checking available hardware") to the old version, and this seems to add a sort of table of contents at the top. Seems nice, why not.

mgeier:
Thanks @cool-RR!

PR #230, opened 2020-03-18
Simplify wheel names

mgeier:
Modern versions of pip should be able to deal with those shorter names.


mgeier:
This might still be problematic nowadays, see https://github.com/pypa/pip/issues/7626.

mgeier:
I'm merging this now, let's see if there are any complaints ...

Issue #128, opened 2018-03-29
Error Getting Mac OS X Sound Output


UsersDoProgram:
I made the following program in an attempt to try and retrieve all the audio that is being played on my Mac and replay it but everytime I run the code in [main.txt](https://github.com/spatialaudio/python-sounddevice/files/1858300/main.txt), I run across the following output,

Default Devices:
```
> 0 Built-in Microphone, Core Audio (2 in, 0 out)
< 1 Built-in Output, Core Audio (0 in, 2 out)
```

```
Traceback (most recent call last):
  File "main.py", line 15, in <module>
    with sd.Stream(channels=2,dtype="float32",callback=callback):
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/sounddevice.py", line 1654, in __init__
    **_remove_self(locals()))
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/sounddevice.py", line 772, in __init__
    'Error opening {0}'.format(self.__class__.__name__))
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/sounddevice.py", line 2563, in _check
    raise PortAudioError(errormsg, err)
sounddevice.PortAudioError: Error opening Stream: Invalid number of channels [PaErrorCode -9998]
```

tgarc:
Hi there! Could you please paste the text of the program or post it in some way that doesn't require downloading? I think this error occurs most commonly when trying to record with a device that is playback only or vice-versa. Without looking at your code, I would suggest specifying ``device=(0,1)`` when instantiating Stream.

mgeier:
@UsersDoProgram You are switching input and output device in your code:

```python
dev = sd.default.device
dev[0],dev[1] = 1,0
sd.default.device = dev
```

The first listed device is the *input* device, the second is the *output*, same order as when you are saying "I/O". You are trying to use the output device for input and vice versa.

If you simply remove everything regarding "device", it should work, because the defaults are already correct.

Or, as @tgarc suggested, you can use `device=(0,1)` in the `sd.Stream()` constructor.

Or, if you want to use `sd.default`, you can try this:

```python
sd.default.device = 0, 1
```

... but this doesn't change anything in your case, since that's already the default.

Maybe the "arrows" in the device list are confusing:

    > 0 Built-in Microphone, Core Audio (2 in, 0 out)
[...]

PR #242, opened 2020-05-04
Separate input and output errors for duplex streams

mgeier:
https://github.com/spatialaudio/python-sounddevice/issues/128#issuecomment-622731448

I'm not convinced that this actually helps.

Instead of the error

```
sounddevice.PortAudioError: Error opening Stream: Invalid number of channels [PaErrorCode -9998]
```

... this would report

```
sounddevice.PortAudioError: Input device error: Invalid number of channels [PaErrorCode -9998]
```

If a user confused the input and output channels, this wouldn't tell them anything more, would it?

I could go a step further and check for the specific error code `-9998` and create a custom message for that (including the requested number of channels), but I'm not sure whether I want to go down that road ...

Any opinions?

### DIFF
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -811,6 +811,13 @@ class _StreamBase(object):
                     'Input and output device must have the same samplerate')
             else:
                 samplerate = isamplerate
+            # Check input and output separately to get more specific errors:
+            _check(
+                _lib.Pa_IsFormatSupported(iparameters, _ffi.NULL, samplerate),
+                'Input device error')
+            _check(
+                _lib.Pa_IsFormatSupported(_ffi.NULL, oparameters, samplerate),
+                'Output device error')
         else:
             parameters, self._dtype, self._samplesize, samplerate = \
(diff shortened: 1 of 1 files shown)


khlorghaal:
I encountered the error because a microphone has 1 channel, and speakers 2, and after being aware of this is obvious, but I was not able to figure out the issue from the error message without googling it.  

If a developer sees they tried getting two channels from an input device, it should be intuitive.

mgeier:
So what exactly are you suggesting?

> If a developer sees they tried getting two channels from an input device, it should be intuitive.

The current error message in this PR would *not* contain the number of channels.

So they wouldn't see that they tried getting two channels from an input device, they would only see that the number of channels for the input device is invalid.

This doesn't tell them that they mixed up the input and output channels.

They still *think* that they provided the right number of channels.

mgeier:
I'm not planning to merge this for now, but if anyone wants to have this (probably in a modified form), please speak up!

Issue #251, opened 2020-06-27
Cannot clear bits in CallbackFlags


mattgwwalker:
The status in a callback provides information on overflow and underflow issues.

Sometimes these issues are acceptable depending on how the callback wants to use the stream.  For example, input underflows and overflows are acceptable in a full-duplex stream if the callback is currently only writing to outdata and thus they could be safely ignored.

It would be useful to be able to clear the bits in the CallbackFlags to indicate the that issues had been considered and deemed acceptable.  At the end of processing in the callback, if any bits were still set, an error could be raised.

However, I see no way to clear specific bits in CallbackFlags.

Would you be open to a Pull Request that added setters for input_underflow, input_overflow, output_underflow, and output_overflow?  Or do you see a better way to implement this?

Cheers,

Matthew

mattgwwalker:
To be more specific, I'm thinking of something along these lines:

![image](https://user-images.githubusercontent.com/4054492/85919595-c8c7f280-b8c0-11ea-8633-e1a7c9b4da69.png)

mgeier:
I've never thought about this.

But for completeness' sake I guess it makes sense to implement this.

> Would you be open to a Pull Request that added setters for input_underflow, input_overflow, output_underflow, and output_overflow?

Yes.

But the docstring will probably not work with Sphinx, we'll have to see about this.

> Or do you see a better way to implement this?

I think the way you suggested is the most straightforward one.

Theoretically, the repetition could be reduced a bit by using a metaclass, but I don't think that's worth it in this case. But I really don't have any experience with metaclasses, so I don't know.

It could probably also be implemented with `__getattr__` (or `__get__` and `__set__`?), but this would be harder to understand and harder to document with Sphinx.

Issue #131, opened 2018-04-06
Is there a way to run doctests?


yarikoptic:
I wondered, since there are docstrings with code samples, and even the `doc/fake__sounddevice.py` may be you guys considered already (so I do not need to fight the windmill) to run the doctests?  On my semi-blind attempt:
```shell
$> python -m nose --with-doctest sounddevice.py             
FFFFF
======================================================================
FAIL: Doctest: sounddevice.AsioSettings.__init__
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3.6/doctest.py", line 2199, in runTest
    raise self.failureException(self.format_failure(new.getvalue()))
AssertionError: Failed doctest test for sounddevice.AsioSettings.__init__
  File "/home/yoh/deb/gits/python-sounddevice/sounddevice.py", line 2093, in __init__

----------------------------------------------------------------------
File "/home/yoh/deb/gits/python-sounddevice/sounddevice.py", line 2117, in sounddevice.AsioSettings.__init__
Failed example:
    sd.play(..., extra_settings=asio_out)
Exception raised:
    Traceback (most recent call last):
      File "/usr/lib/python3.6/doctest.py", line 1330, in __run
        compileflags, 1), test.globs)
      File "<doctest sounddevice.AsioSettings.__init__[2]>", line 1, in <module>
        sd.play(..., extra_settings=asio_out)
      File "/home/yoh/deb/gits/python-sounddevice/sounddevice.py", line 135, in play
        ctx.frames = ctx.check_data(data, mapping, kwargs.get('device'))
      File "/home/yoh/deb/gits/python-sounddevice/sounddevice.py", line 2306, in check_data
        dtype = _check_dtype(data.dtype)
      File "/home/yoh/deb/gits/python-sounddevice/sounddevice.py", line 2459, in _check_dtype
        raise TypeError('Unsupported data type: ' + repr(dtype))
    TypeError: Unsupported data type: 'object'
----------------------------------------------------------------------
File "/home/yoh/deb/gits/python-sounddevice/sounddevice.py", line 2122, in sounddevice.AsioSettings.__init__
Failed example:
    sd.play(...)
Exception raised:
    Traceback (most recent call last):
      File "/usr/lib/python3.6/doctest.py", line 1330, in __run
        compileflags, 1), test.globs)
      File "<doctest sounddevice.AsioSettings.__init__[4]>", line 1, in <module>
        sd.play(...)
[...]

Issue #14, opened 2015-12-09
Implement PortAudio loopback test in Python


mgeier:
See https://subversion.assembla.com/svn/portaudio/portaudio/trunk/qa/loopback/README.txt

With a bit of luck this can be used with the JACK dummy backend as an automated test on http://travis-ci.org ...

mgeier:
Git link: https://app.assembla.com/spaces/portaudio/git/source/master/qa/loopback

tgarc:
It seems like porting this would substantially increase the size of sounddevice's codebase. It also seems like, while they are useful tests, porting them wouldn't really serve the purpose of testing the functionality of sounddevice. Thoughts?

mgeier:
Yes, to port the whole thing 1:1 is probably too much.
But some subset might still be interesting.
It doesn't have to be the exact functionality from PortAudio either, but I guess that could serve as an inspiration for some kinds of similar tests.

My goal with that would be twosome:

* for users to be able to quickly check if the advertised latency values are accurate on their system. I guess this very much depends on the host API and the hardware in use, therefore users should be able to run this without much hassle on their machines whenever they are curious. They could probably also run the original test programs from PortAudio, but those might be hard to get on their platform.

* fully automated tests that can run on a CI server without user interaction (and without any hardware devices present). This would of course only need to test the functionality of the `sounddevice` module, it doesn't need to test PortAudio iself. It doesn't need to test all the host APIs, just a single one with a virtual loopback cable (JACK `dummy` driver, ALSA with `snd-aloop`, ...?) would be enough.

It definitely doesn't have to cover the whole scope of the PortAudio tests. Especially the whole auto-detection mechanism of loopback cables isn't necessary since we can just specify which device and which channels we want to test.

It might even make sense to encapsulate the signal generation and detection algorithms into a separate project, since this really isn't the job of the `sounddevice` module. And it could theoretically be re-used by other Python libraries.

Or probably there are already libraries available that do all this stuff?

Would you be interested in working on that?

tgarc:
It might me interesting to implement separately..I'd have to spend time exploring that code, I've only just skimmed over it so far. Certainly if you make any progress you should make a branch or another repository so that I could play with it as well.

I think some functionality for latency testing would be really interesting! Probably wouldn't be that difficult to write either. I think it would be useful for users too.

In the short term though, I think the priority would be to add a testing module for sounddevice that tests some basic functionality and loopback tests that work across platforms.

[...]

PR #255, opened 2020-07-03
Added setters to CallbackFlag's bits

mattgwwalker:
Hi,

This Pull Request adds the setters to CallbackFlags as we discusses in Issue #251.

I wasn't able to fathom out how to get PyTest working with sounddevice.  I had issues with the ffi import.  Had I been able to get it to work, I would have added tests to demonstrate the correct functionality.  As it stands, I just did the testing manually:

Screen Shot 2020-07-04 at 10 10 18

Finally, a comment about the instructions in the file "CONTRIBUTING".  Because I am working in a virtual environment, I found a variation of the pip install command worked best for me:
    pip install -e .

Would you like that added to the documentation?

Cheers and thanks for all your work on sounddevice!

Matthew

### DIFF
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -1643,12 +1643,6 @@ class Stream(InputStream, OutputStream):
             next practical value -- i.e. to provide an equal or higher
             latency  wherever possible.  Actual latency values for an
             open stream may be retrieved using the `latency` attribute.
-
-            .. note:: Specifying the desired latency as 'high' does
-                not guarantee a stable audio stream. For reference, by
-                default Audacity specifies a desired latency of 100ms and
-                achieves robust performance.
-
         extra_settings : settings object or pair thereof, optional
             This can be used for host-API-specific input/output
             settings.  See `default.extra_settings`.
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -1972,12 +1972,7 @@ class CallbackFlags(object):
         return bool(self._flags & flag)
 
     def _updateflag(self, flag, value):
-        """Set/reset a given flag.
-        
-        Set the given flag if value is True, otherwise clear the
-        given flag.
-        
-        """
+        """Set/clear a given flag."""
         if value:
             self._flags |= flag
         else:
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -1643,6 +1643,12 @@ class Stream(InputStream, OutputStream):
             next practical value -- i.e. to provide an equal or higher
             latency  wherever possible.  Actual latency values for an
             open stream may be retrieved using the `latency` attribute.
+
+            .. note:: Specifying the desired latency as 'high' does
+                not guarantee a stable audio stream. For reference, by
+                default Audacity specifies a desired latency of 100ms and
+                achieves robust performance.
+
         extra_settings : settings object or pair thereof, optional
             This can be used for host-API-specific input/output
             settings.  See `default.extra_settings`.
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -1972,8 +1972,10 @@ class CallbackFlags(object):
         return bool(self._flags & flag)
 
     def _updateflag(self, flag, value):
-        """Set the given flag if value is True, otherwise clear the given
-        flag."""
+        """Set/reset a given flag.
+        
+        Set the given flag if value is True, otherwise clear the
+        given flag."""
         if value:
             self._flags |= flag
         else:
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -1838,6 +1838,19 @@ class CallbackFlags(object):
     >>> errors.input_overflow
     True
 
+    The values may also be set and cleared by the user:
+
+    >>> import sounddevice as sd
+    >>> cf = sd.CallbackFlags()
+    >>> cf
+    <sounddevice.CallbackFlags: no flags set>
+    >>> cf.input_underflow = True
+    >>> cf
+    <sounddevice.CallbackFlags: input underflow>
+    >>> cf.input_underflow = False
+    >>> cf
+    <sounddevice.CallbackFlags: no flags set>
+
     """
 
     __slots__ = '_flags'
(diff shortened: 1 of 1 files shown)


mattgwwalker:
I didn't say: as for the docstring, I was interested to learn that the setters should be documented in the getter [1].  I never expected that!

[1] https://stackoverflow.com/questions/17478040/how-can-i-generate-documentation-for-a-python-property-setter-using-sphinx

sounddevice.py
```diff
@@ -1877,9 +1877,19 @@ def input_underflow(self):
         This can only happen in full-duplex streams (including
         `playrec()`).
 
+        The bit may be set with ``input_underflow = True``, and
+        cleared with ``input_underflow = False``.
```
mgeier, line 1881: I don't really like this documentation.

I don't really like to repeat this for every field, because I don't think its that important for most people.

More importantly, it's not working, you would have to have something like `mystatus.input_underflow = True`.

What about adding a second example to the already existing example:

https://github.com/spatialaudio/python-sounddevice/blob/5c1a9de051cfa6b72bf26923c37a43c5ee5a0a59/sounddevice.py#L1828-L1840

I don't think we have to repeat this for every single field name, our users should be able to infer that it works for every field, not only the ones that happen to be used in the example.

sounddevice.py
```diff
@@ -1877,9 +1877,19 @@ def input_underflow(self):
...
+
         """
         return self._hasflag(_lib.paInputUnderflow)
 
+    @input_underflow.setter
+    def input_underflow(self, new_value):
```
mgeier, line 1887: I don't think the `new_` part of `new_value` is necessary. I would prefer `value`.

See also https://docs.python.org/3/library/functions.html#property.

sounddevice.py
```diff
@@ -1893,9 +1903,19 @@ def input_overflow(self):
...
+
         """
         return self._hasflag(_lib.paInputOverflow)
 
+    @input_overflow.setter
+    def input_overflow(self, new_value):
+        if new_value:
+            self._setflag(_lib.paInputOverflow)
+        else:
+            self._clearflag(_lib.paInputOverflow)
```
mgeier, line 1917: I guess it doesn't make a huge difference, but you are repeating the `if` statement for each flag.

What about moving it to the helper function?

```python
    @input_overflow.setter
    def input_overflow(self, value):
        self._updateflag(_lib.paInputOverflow, value)

    ...

    def _updateflag(self, flag, value):
        if value:
            self._flags |= flag
        else:
            self._flags &= ~flag
```

mgeier:
> I wasn't able to fathom out how to get PyTest working with sounddevice. I had issues with the ffi import. Had I been able to get it to work, I would have added tests to demonstrate the correct functionality.

Yeah, the testing situation isn't great, I admit.

There is already #131 which talks about testing and problems with CFFI. Is yours the same problem?
Could you please create a separate issue for this problem?

The main question would be if we actually want to load the PortAudio library for testing or not?

This might cause problems on CI servers where no sound card is accessible.

See also #14 for ideas for running tests without physical hardware.

I would be happy for any improvements regarding testing!

> Finally, a comment about the instructions in the file "CONTRIBUTING". Because I am working in a virtual environment, I found a variation of the pip install command worked best for me:
> pip install -e .

You mean that the `--user` option I've mentioned in

https://github.com/spatialaudio/python-sounddevice/blob/5c1a9de051cfa6b72bf26923c37a43c5ee5a0a59/CONTRIBUTING.rst#L95

... is misleading?

I've described this flag in https://python-sounddevice.readthedocs.io/en/0.3.15/installation.html, but we can repeat that in `CONTRIBUTING.rst`.

Or is it something else?

> Would you like that added to the documentation?

Sure!

sounddevice.py
mattgwwalker: I feel embarrassed I didn't see this myself!  Thanks :o)

sounddevice.py
mattgwwalker: I have removed the previous documentation and replaced it with an example:

    The values may also be set and cleared by the user:                                                                           
                                                                                                                                  
    >>> import sounddevice as sd                                                                                                  
    >>> cf = sd.CallbackFlags()                                                                                                   
    >>> cf                                                                                                                        
    <sounddevice.CallbackFlags: no flags set>                                                                                     
    >>> cf.input_underflow = True                                                                                                 
    >>> cf                                                                                                                        
    <sounddevice.CallbackFlags: input underflow>                                                                                  
    >>> cf.input_underflow = False                                                                                                
    >>> cf                                                                                                                        
    <sounddevice.CallbackFlags: no flags set>

sounddevice.py
mattgwwalker: I've removed the "new_" part from each setter.

mattgwwalker:
Thanks very much for such a detailed review.  I think your review was really gold-standard; actually, I think it's the best I've seen.

Hopefully these changes are acceptable.

Cheers,

Matthew

mattgwwalker:
Oh, also, as previously, I hand tested the changes:

image

sounddevice.py
```diff
@@ -1942,6 +1971,14 @@ def _hasflag(self, flag):
         """Check a given flag."""
         return bool(self._flags & flag)
 
+    def _updateflag(self, flag, value):
+        """Set the given flag if value is True, otherwise clear the given
+        flag."""
```
mgeier, line 1976: The "summary" should fit into a single line.

Probably you can shorten it?

What about "Set/reset a given flag."?

This is what `pydocstyle` has to say:

```
sounddevice.py:1974 in private method `_updateflag`:
        D205: 1 blank line required between summary line and description (found 0)
sounddevice.py:1974 in private method `_updateflag`:
        D209: Multi-line docstring closing quotes should be on a separate line
sounddevice.py:1974 in private method `_updateflag`:
        D400: First line should end with a period (not 'n')
```

sounddevice.py
mgeier: No need to be embarrassed.

That's what reviews are for. It's always good to have a second (or third etc.) set of eyes looking at the code.

mgeier:
Thanks for the update!

>  I think your review was really gold-standard; actually, I think it's the best I've seen.

Wow, thanks.

Since this codebase is very small and PRs are kinda rare, I can take the time to go into the details.

A bigger project probably wouldn't have this luxury.

> Hopefully these changes are acceptable.

Certainly, except for the little nit I mentioned above.

mgeier:
@mattgwwalker Can you please fix the docstring I've mentioned above (https://github.com/spatialaudio/python-sounddevice/pull/255#discussion_r455120802)?

Then I can merge this PR.

sounddevice.py
mattgwwalker: Thanks for this feedback.  I learnt something about docstrings that I didn't know!

Issue #262, opened 2020-07-09
Desired Latency


mattgwwalker:
Hi,

I have been struggling with this issue for weeks, and yesterday found (the now obvious) solution.

I was regularly receiving input_underflows when connecting an HDMI output to a Mac in a full-duplex stream.  Sometimes dozens of the errors, mostly they occurred at the beginning of the stream.

I was passing "high" as the desired latency to the Stream constructor.  I believed that this would be well-defined to always provide sufficient buffering to ensure that the stream was stable.  Golly me, how wrong was I?!

I dug into Audacity's code to see how they were achieving such stable performance on the same hardware and all that I could see they were doing was specifying a specific desired latency, which defaults to 100ms.

By copying their approach (dropping the use of "high" and specifying 100ms) I now have very much more robust audio.

Hopefully this comment will help someone else not waste weeks chasing their tail!

As an aside, asking for 100ms seems to be very very much a lower limit (at least on this Mac).  My measurements of the total latency with an HDMI output (so from the time it's asked to play to the time it hears itself through the microphone) were 285ms +/- 2ms.  The total latency versus the desired latency specified to the Stream constructor is also very non-linear:

![image](https://user-images.githubusercontent.com/4054492/87091642-24658900-c28e-11ea-9890-0f0c0f6aa044.png)

Given my frustrating experience, perhaps it's worthwhile adding something into the documentation?  How would you feel if I added the following into the section "latency" under Stream[1]:

- Specifying the desired latency as "high" does not guarantee a stable audio stream.  For reference, by default Audacity specifies a desired latency of 100ms and achieves robust performance.

[1] https://python-sounddevice.readthedocs.io/en/0.3.15/api/streams.html#sounddevice.Stream

mgeier:
There's already a short hint in https://python-sounddevice.readthedocs.io/en/latest/api/module-defaults.html#sounddevice.default.latency:

> `'high'` is typically more robust (i.e. buffer under-/overflows are less likely)

But it would be great if you could make this more clear in the docs!

What is `desired_latency` in your plot?

Is it the `latency` you requested from the stream or is it `stream.latency`?

The `Stream` docs say (mostly copied from the Portaudio docs):

[...]

PR #263, opened 2020-07-13
High latency doc

mattgwwalker:
Hi,

I added a note to the documentation of a Stream's desired latency indicating that 'high' may not be sufficient to obtain stable audio.

In my commit history there's a revert.  Does this impact you in any way?  Or do my commits just get merged into one at your end?

Cheers,

Matthew

### DIFF
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -1883,18 +1883,8 @@ class CallbackFlags(object):
         This can only happen in full-duplex streams (including
         `playrec()`).
 
-        The bit may be set with ``input_underflow = True``, and
-        cleared with ``input_underflow = False``.
-
         """
         return self._hasflag(_lib.paInputUnderflow)
-
-    @input_underflow.setter
-    def input_underflow(self, new_value):
-        if new_value:
-            self._setflag(_lib.paInputUnderflow)
-        else:
-            self._clearflag(_lib.paInputUnderflow)
 
     @property
     def input_overflow(self):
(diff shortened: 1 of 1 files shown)


mattgwwalker:
Oh, for reference, we discussed this in #262.

sounddevice.py
```diff
@@ -1643,6 +1643,12 @@ def __init__(self, samplerate=None, blocksize=None,
             next practical value -- i.e. to provide an equal or higher
             latency  wherever possible.  Actual latency values for an
             open stream may be retrieved using the `latency` attribute.
+
+            .. note:: Specifying the desired latency as 'high' does
+                not guarantee a stable audio stream. For reference, by
```
mgeier, line 1648: I don't think it is obvious to most readers what exactly "stable" means here?

mgeier:
Thanks for this PR!

I think I would like this to be a larger change.

This sentence is copied from the PortAudio docs, but it is quite useless in the `sounddevice` context:

            Where practical, implementations should configure their
            latency based on this parameter, otherwise they may choose
            the closest viable latency instead.  Unless the suggested
            latency is greater than the absolute upper limit for the
            device, implementations should round the *latency* up to the
            next practical value -- i.e. to provide an equal or higher
            latency  wherever possible.

Isn't it?

Even if the information is useful, it's written from the wrong point of view.

Would you like to fix this?

Also, I don't like that there is divergent information between the `latency` argument to the `Stream` constructor and [sd.default.latency](https://python-sounddevice.readthedocs.io/en/0.3.15/api/module-defaults.html#sounddevice.default.latency).

I think I'd prefer to document only one and link to it from the other one.

That's also what I've done with [sd.default.blocksize](https://python-sounddevice.readthedocs.io/en/0.3.15/api/module-defaults.html#sounddevice.default.blocksize).

mgeier:
> In my commit history there's a revert. Does this impact you in any way? Or do my commits just get merged into one at your end?

No problem, I can squash the commits in the end.

PR #264, opened 2020-07-13
Added comments on installation under a virtual environment

mattgwwalker:
Added comments on how to install sounddevice when in a virtual environment (discussed in #255).

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -100,6 +100,12 @@ newest development version (a.k.a. "master") from Github_::
 This way, your installation always stays up-to-date, even if you pull new
 changes from the Github repository.
 
+If you're inside a Python `virtual environment`_, replace the last line with::
+
+   pip install -e .
+
+.. _`virtual environment`: https://docs.python.org/3/tutorial/venv.html
+
 Whenever the file ``sounddevice_build.py`` changes (either because you edited it
 or it was updated by pulling from Github or switching branches), you have to run
 the last command again.
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -54,6 +54,10 @@ To install the latest release from PyPI, use::
 
    python3 -m pip install sounddevice --user
 
+Or, if you're inside a Python `virtual environment`_::
+
+   pip install sounddevice
+   
 If you want to install it system-wide for all users (assuming you have the
 necessary rights), you can just drop the ``--user`` option.
 If you have installed the module already, you can use the ``--upgrade`` flag to
@@ -89,3 +93,4 @@ might require a compiler and several additional libraries.
 .. _WinPython: https://winpython.github.io/
 .. _CFFI: https://cffi.readthedocs.io/
 .. _PyPI: https://pypi.org/project/sounddevice/
+.. _`virtual environment`: https://docs.python.org/3/tutorial/venv.html


mgeier:
You are mentioning two changes:

* `pip` instead of `python3 -m pip`
* `--user` vs. no `--user`

Are both needed?

I understand the second, but I'm not sure why the first is needed.

When using only `pip`, you normally don't know which Python version is used. I like to explicitly mention the Python executable.

Also, there are similar instructions in on the page https://python-sounddevice.readthedocs.io/en/latest/installation.html. Should those be changed as well?

I think they should either be both the same, or the main installation page should have all the details and the "contributing" page has only the basic case, referring to the main installation page.

What would you prefer?

I'm trying to make/keep this consistent over multiple projects, so once a good version is found here, I will update several other pages:

* https://jackclient-python.readthedocs.io/en/0.5.2/#installation
* https://jackclient-python.readthedocs.io/en/0.5.2/#contributing
* https://nbsphinx.readthedocs.io/en/0.7.1/installation.html
* https://nbsphinx.readthedocs.io/en/0.7.1/contributing.html
* https://jupyter-format.readthedocs.io/en/latest/installation.html
* https://sfs-python.readthedocs.io/en/0.5.0/installation.html
* https://sfs-python.readthedocs.io/en/0.5.0/contributing.html

mattgwwalker:
If you've created a virtual environment, then you **do** know what version of python and pip will be called.  They're part of the specification of the virtual environment.

```
~$ which python
/usr/bin/python
~$ python --version
Python 2.7.16
~$ which pip
/usr/local/bin/pip
~$ pip --version
pip 18.1 from /Library/Python/2.7/site-packages/pip (python 2.7)
~$ python3 -m venv my_virtual_environment
~$ source my_virtual_environment/bin/activate
(my_virtual_environment) ~$ which python
/Users/matthew/my_virtual_environment/bin/python
(my_virtual_environment) ~$ python --version
Python 3.7.7
(my_virtual_environment) ~$ which pip
/Users/matthew/my_virtual_environment/bin/pip
(my_virtual_environment) ~$ pip --version
pip 19.2.3 from /Users/matthew/my_virtual_environment/lib/python3.7/site-packages/pip (python 3.7)
```

If you upgrade pip, for example, it will upgrade in the virtual environment only.  If you install modules, they will be installed only in the virtual environment.

This gives you the advantage of being able to have multiple different configurations on one machine, one per project.

Calling another version of python (e.g. python2.7) from within the virtual environment would be certainly going against the intended grain, but I can't see that they discuss it in the [official docs](https://docs.python.org/3/tutorial/venv.html).

So, all that to say, yes, I think the two changes are both required :o)

You're right about the installation documentation.  I didn't consider that.  Users too might prefer to install into a virtual environment.

mgeier:
Thanks for the update!

What happens when you do the following?

    (my_virtual_environment) ~$ which python3

mattgwwalker:
```
~$ which python
/usr/bin/python
~$ which python3
/usr/local/bin/python3
~$ python3 -m venv my_virtual_environment
~$ source my_virtual_environment/bin/activate
(my_virtual_environment) ~$ which python
/Users/matthew/my_virtual_environment/bin/python
(my_virtual_environment) ~$ which python3
/Users/matthew/my_virtual_environment/bin/python3
(my_virtual_environment) ~$ ls -l my_virtual_environment/bin/python
lrwxr-xr-x  1 matthew  staff     7B 17 Jul 09:27 my_virtual_environment/bin/python@ -> python3
(my_virtual_environment) ~$ ls -l my_virtual_environment/bin/python3
lrwxr-xr-x  1 matthew  staff    22B 17 Jul 09:27 my_virtual_environment/bin/python3@ -> /usr/local/bin/python3
```

mattgwwalker:
You can create virtual environments this way from Python 3.3+

The Python standard library version is a subset of `virtualenv`, which works from Python 2.7+.  See https://virtualenv.pypa.io/en/latest/

mgeier:
OK, so this should work in your virtual env:

    python3 -m pip install -e .

So only the `--user` flag has to be removed. BTW, the error message if you fail to do this is quite helpful:

    ERROR: Can not perform a '--user' install. User site-packages are not visible in this virtualenv.

I would use a text somewhat like this:

```rst
Depending on your Python installation,
you may have to use ``python`` instead of ``python3``.
If you want to install the module system-wide for all users
(assuming you have the necessary rights),
you should drop the ``--user`` flag.
If you are using a virtual environment,
you should also drop the ``--user`` flag.
If you have installed the module already,
you can use the ``--upgrade`` flag to get the newest release.
```

Do you think this should be used for both "installation" and "contributing" or only the former?

In the "contributing" text, I would probably use something like this:

```rst
... where ``-e`` stands for ``--editable``.
This way, your installation always stays up-to-date,
even if you pull new changes from the Github repository.
```

And here is a possible text for some Git instructions for "contributing":

```rst
When installing this way, you can quickly try other Git
branches (in this example the branch is called "another-branch")::

   git checkout another-branch

If you want to go back to the "master" branch, use::

   git checkout master

To get the latest changes from Github, use::

   git pull --ff-only
```

mattgwwalker:
So, I had a shower thought the other day: if `python4` were to exist (as it most probably will one day) and a user created a virtual environment with it, then specifying `python3` in the docs would potentially send users up the boohai.

This can be seen by imagining the docs specified python2.7 in a python3 world:

```
~$ python3 -m venv my-virtual-env
~$ source my-virtual-env/bin/activate
(my-virtual-env) ~$ which python2.7
/opt/local/bin/python2.7
(my-virtual-env) ~$ which python
/Users/matthew/my-virtual-env/bin/python

```

mattgwwalker:
> So only the --user flag has to be removed. BTW, the error message if you fail to do this is quite helpful:
> `ERROR: Can not perform a '--user' install. User site-packages are not visible in this virtualenv.`

I think including the error message would be a bloody greatly idea actually.  Good thinking.  Most users will probably do their thing without reading the docs first (I might be one of the guilty party here) and a fair portion of them may Google the error message expecting Google to just point them to the correct part of the documentation.  So including it in the docs would indeed be a good thing.

mattgwwalker:
On the subject of virtual environments, I've been using `pipenv` recently.  I especially like the `pipenv shell` command that lets you drop into the correct virtual environment from wherever you are in the source tree.

mattgwwalker:
> ... where ``-e`` stands for ``--editable``.
> This way, your installation always stays up-to-date,
> even when you pull new changes from the Github repository.

I've also found the editable flag extremely useful.  However, it's also caused me incredible hours of wasted time  just recently as it appears PyInstaller is silently incompatible with libraries installed in this way.  PEX too is incompatible with the `editable` flag, but at least it's clear that it wants pre-packaged distributions.

mgeier:
> I think including the error message would be a bloody greatly idea actually. Good thinking.

Something like this perhaps?

```rst
If you are using a `virtual environment`_,
you should also drop the ``--user`` flag
(otherwise you'll get an error like
``ERROR: Can not perform a '--user' install.
User site-packages are not visible in this virtualenv.``).
```

> On the subject of virtual environments, I've been using `pipenv` recently. I especially like the `pipenv shell` command that lets you drop into the correct virtual environment from wherever you are in the source tree.

Sounds good, but do you think this should be mentioned in the `sounddevice` docs?

> I've also found the editable flag extremely useful. However, it's also caused me incredible hours of wasted time just recently as it appears PyInstaller is silently incompatible with libraries installed in this way. PEX too is incompatible with the `editable` flag, but at least it's clear that it wants pre-packaged distributions.

OK, good to know. But what does that mean for the `sounddevice` docs?

mattgwwalker:
> If you are using a `virtual environment`_,
you should also drop the ``--user`` flag
(otherwise you'll get an error like
``ERROR: Can not perform a '--user' install.
User site-packages are not visible in this virtualenv.``).

Sounds good to me.



>> On the subject of virtual environments, I've been using pipenv recently. I especially like the pipenv shell command that lets you drop into the correct virtual environment from wherever you are in the source tree.

> Sounds good, but do you think this should be mentioned in the sounddevice docs?

No, I don't think a specific virtual environment should be mentioned in the docs.




>> I've also found the editable flag extremely useful. However, it's also caused me incredible hours of wasted time just recently as it appears PyInstaller is silently incompatible with libraries installed in this way. PEX too is incompatible with the editable flag, but at least it's clear that it wants pre-packaged distributions.

> OK, good to know. But what does that mean for the sounddevice docs?

I wish I knew how better to manage the development environment.  I don't have any gold-standard methods that I'd recommend enough for the docs.

mgeier:
OK, thanks.

If you want to update this PR, feel free to do so.

mgeier:
I've been told that modern `pip` versions don't need the `--user` flag anymore (https://github.com/spatialaudio/python-rtmixer/pull/36#pullrequestreview-476936162), so I'm suggesting to remove it in #277.

PR #268, opened 2020-07-27
Fix exception causes all over the codebase

cool-RR:
Now that Python 2 support has been dropped, we can use `raise foo from bar`

More details about this change: https://blog.ram.rachum.com/post/621791438475296768/improving-python-exception-chaining-with

### DIFF
--- a/examples/play_long_file.py
+++ b/examples/play_long_file.py
@@ -72,9 +72,9 @@ def callback(outdata, frames, time, status):
     assert not status
     try:
         data = q.get_nowait()
-    except queue.Empty:
+    except queue.Empty as e:
         print('Buffer is empty: increase buffersize?', file=sys.stderr)
-        raise sd.CallbackAbort
+        raise sd.CallbackAbort from e
     if len(data) < len(outdata):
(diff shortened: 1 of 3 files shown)


examples/play_long_file.py
```diff
@@ -72,9 +72,9 @@ def callback(outdata, frames, time, status):
     assert not status
     try:
         data = q.get_nowait()
-    except queue.Empty:
+    except queue.Empty as e:
         print('Buffer is empty: increase buffersize?', file=sys.stderr)
-        raise sd.CallbackAbort
+        raise sd.CallbackAbort from e
```
mgeier, line 77: This will never bubble up to the user, see:

https://github.com/spatialaudio/python-sounddevice/blob/afe0a822b5591c2e3ae7fe9f1517d7342e048c46/sounddevice.py#L2648-L2651

But I guess it still doesn't hurt to add the `from e`, right?

cool-RR: Yes, I think in this case it's better to stay correct even when there aren't any practical implications.

Issue #279, opened 2020-09-07
How to specify a device from a query?


ullix:
On my Ubuntu 16.04 computer a call to `sd.query_devices()` gives this answer (some lines removed):

> 4 HDA Intel PCH: ALC887-VD Analog (hw:1,0), ALSA (2 in, 8 out)
5 HDA Intel PCH: ALC887-VD Digital (hw:1,1), ALSA (0 in, 2 out)
6 HDA Intel PCH: ALC887-VD Alt Analog (hw:1,2), ALSA (2 in, 0 out)
    *9 default, ALSA (32 in, 32 out)

When I then call the # 4 device by `print(sd.query_devices(device="HDA Intel PCH: ALC887-VD Analog", kind= 'input'))` I get an exception:

> ValueError: Multiple input devices found for 'HDA Intel PCH: ALC887-VD Analog':
[4] HDA Intel PCH: ALC887-VD Analog (hw:1,0), ALSA
[6] HDA Intel PCH: ALC887-VD Alt Analog (hw:1,2), ALSA

So how can I then select # 4 when this cannot be distinguished from # 6, although the 'Alt' word is missing in the query?

Selecting # 6 with `device="HDA Intel PCH: ALC887-VD Alt Analog"` does work. Once this is done, I can call the non-Alt # 4 device, and not have an exception, but: this works because the command automatically calls the # 6 device, and not # 4!

So, no way to select # 4. What else can I provide when using the default names?

mgeier:
The string is not compared "exactly", see https://python-sounddevice.readthedocs.io/en/0.4.0/api/module-defaults.html#sounddevice.default.device:

> If a string is given, the device is selected which contains all space-separated parts in the right order. Each device string contains the name of the corresponding host API in the end. The string comparison is case-insensitive.

In your case, you should be able to use `device='hw:1,0'` or `device=4`.

raveslave:
how about adding / exposing the serial-number from the raw usb descriptor itself?

this way each sound-card would have a unique handle that can also be used to create a more flexible connectionID that doesn't shift (when things are hotplugged) as with an index based setup.

mgeier:
@raveslave

> how about adding / exposing the serial-number from the raw usb descriptor itself?
> 
> this way each sound-card would have a unique handle that can also be used to create a more flexible connectionID that doesn't shift (when things are hotplugged) as with an index based setup.

[...]

PR #360, opened 2021-07-23
Allow exact matches to include the hostapi's name in _get_device_id

blattm:
Hi there,

here is a suggestion for a change of _get_device_id:
When a string of the form "device_name, host_api" name is queried, it should count as an exact match if it matches full_string.

This would solve the following problem, I currently encounter. I want to use strings in the specified format to store audio settings in configuration files. But this will happen:
```
>>> sd.query_devices("default, ALSA")
ValueError: Multiple input/output devices found for 'default, ALSA':
[6] sysdefault, ALSA
[14] default, ALSA
```
Without the hostapi part, the error does not occur. However, I cannot omit it, because on Windows, there are sometimes devices with identical names, but different APIs.

### DIFF
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -2739,7 +2739,7 @@ def _check(err, msg=''):
 
 
 def _remove_alsa_id(device_string):
-    return _re.sub(" \(hw:[0-9]+,[0-9]+\)$", "", device_string)
+    return _re.sub(r" \(hw:[0-9]+,[0-9]+\)$", "", device_string)
 
 
 def _get_device_id(id_or_query_string, kind, raise_on_error=False):
@@ -2784,7 +2784,10 @@ def _get_device_id(id_or_query_string, kind, raise_on_error=False):
             pos += len(substring)
         else:
             matches.append((id, full_string))
-            if query_string[:len(device_string)].lower() == device_string.lower():
+            if (
+                query_string[: len(device_string)].lower()
+                == device_string.lower()
+            ):
                 exact_device_matches.append(id)
 
     if kind is None:
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -53,6 +53,7 @@ __version__ = '0.4.2'
 import atexit as _atexit
 import os as _os
 import platform as _platform
+import re as _re
 import sys as _sys
 from ctypes.util import find_library as _find_library
 from _sounddevice import ffi as _ffi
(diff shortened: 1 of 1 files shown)


blattm:
I just realized that the way I planned to craft identifying strings for config files is problematic.
There are two reasons:
1) The names of the hostapis are not stable, if I correctly understand this comment https://github.com/spatialaudio/python-sounddevice/issues/84#issuecomment-297292987 . However, I don't think this is as much of a problem, because such changes will probably not happen that often.
2) ALSA device names of low-level hardware include a string such as `"hw:1,0"`. Querying without that string might not work ( #279 ) and querying the full name including `"hw:1,0"` might not work, as these numbers could potentially change if external hardware gets connected in a different order.

So I'm not sure if this pull request would really help in its current form...

One possibility, that would also help with 2. and #279: What if we exclude strings like `" (hw:1,0)"` from the `device_string`, but still include it in the `full_string` for ALSA-devices. But this would not help if `query_string` contained the hostapi... So maybe one could introduce an additional parameter like `hostapi=None` in `_get_device_id` and `query_devices`..?
Just bouncing ideas.

What do you think?

blattm:
Ok, I think I found a nice solution, that will not introduce changes to the public API, so I updated the PR :)

The new code would allow the exact matching of the device's name, if
1) the hostapi is not part of the query string, 
2) the exact hostapi's name is at the end of the query string, or
3) a part of the hostapi's name is at the end of the query string

(The current implementation of 0.4.2 allows 1., the initial PR only allows 1. and 2.)

Also it solves #279, even when the query contains the hostapi's name.

With this change, it will be possible to craft strings for config files, that will **always** give you the correct device when queried, like this: `sd._remove_alsa_id(device_name) + " " + hostapi_name`

mgeier:
Thanks for this PR!

> When a string of the form "device_name, host_api" name is queried, it should count as an exact match if it matches full_string.

Yes, that sounds reasonable, I wasn't aware of this problem.

I don't really like the recent update, though.

I don't want to make any assumptions about the device string that's provided by PortAudio.
If you want to do that, you should do that in your own code.

mgeier:
I would like to merge the change in f1fcaaf387e22e2470254515a5c09e91dce0bebf, but I'm not convinced of the following commits.

@blattm I assume the first commit would already be helpful to you, right?

Are you OK with me just taking the first commit?

blattm:
Hi,

I'm fine with that!

> I don't want to make any assumptions about the device string that's provided by PortAudio. If you want to do that, you should do that in your own code.

I thought that was a reasonable boundary to draw, so I already solved the problem within my own code.

mgeier:
Thanks!

I've cherry-picked your first commit into the `master` branch as a6248d50f3f6e99bcc10fad73f2489f13c6a39f2.

PR #396, opened 2022-03-05
add GitHub URL for PyPi

andriyor:
Warehouse now uses the project_urls provided to display links in the sidebar on [this screen](https://pypi.org/project/requests/), as well as including them in API responses to help the automation tool find the source code for Requests.

Example: [pypi.org/pypi/requests/json](https://pypi.org/pypi/requests/json)
Docs: [packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -77,6 +77,9 @@ setup(
     license='MIT',
     keywords='sound audio PortAudio play record playrec'.split(),
     url='http://python-sounddevice.readthedocs.io/',
+    project_urls={
+        'Source': 'https://github.com/spatialaudio/python-sounddevice',
+    },
     platforms='any',
     classifiers=[
         'License :: OSI Approved :: MIT License',


mgeier:
Thanks! I wanted to add this at some point, but I must have forgotten.

What about changing `Source` to `Source Code`? I think this would be clearer.

And what about adding `Bug Tracker`?

Another example:
https://pypi.org/project/nbsphinx/
https://github.com/spatialaudio/nbsphinx/blob/d7241b111178fb850f09b52574c8d1d720c84099/setup.py#L31-L35

Interestingly, in your JSON example, the `bugtrack_url` and `docs_url` are empty, do you know what's up with that?

andriyor:
`Source` is the most used key https://github.com/pypa/warehouse/issues/5947

mgeier:
"most used" doesn't necessarily mean "best choice". But if you prefer `Source`, that's fine with me.

Thanks again!

Issue #11, opened 2015-11-09
Hide messages from PortAudio on stdout/stderr


mgeier:
Mostly, those messages are annoying, but I don't know if they ever show useful information.

We could try to redirect the standard output from PortAudio to `os.devnull`, as shown here:

http://eli.thegreenplace.net/2015/redirecting-all-kinds-of-stdout-in-python/
http://stackoverflow.com/questions/5081657/how-do-i-prevent-a-c-shared-library-to-print-on-stdout-in-python

mgeier:
I didn't find a way to redirect the output for all PortAudio calls but still leave `stderr` working for other uses within Python.

It seems to work by running this before each PortAudio call:

```python
devnull = os.open(os.devnull, os.O_WRONLY)
old_stderr = os.dup(2)
sys.stderr.flush()
os.dup2(devnull, 2)
os.close(devnull)
```

... and this immediately afterwards (ideally, in a with statement):

```python
os.dup2(old_stderr, 2)
os.close(old_stderr)
```

However, I'm not sure if I want to add this overhead to every single PortAudio call.

Any comments?

mgeier:
I found a different way, but this doesn't work on Windows (and probably Solaris), because there is no symbol named `stderr`.

On Linux, this works fine:

Add this to the `_ffi.cdef()` call:

```c
/* from stdio.h */
FILE* fopen(const char* path, const char* mode);
FILE* stderr;
```

And then add this just before loading the PortAudio library:

```python
# Ignore stderr messages from PortAudio
_cstdlib = _ffi.dlopen(None)
_cstdlib.stderr = _cstdlib.fopen(_os.devnull.encode(), b'w')
```

This directly overwrites the `stderr` file stream, which doesn't affect Python's `sys.stderr` at all.
Therefore, there is no need to undo the redirection in the end.

Issue #10, opened 2015-11-07
Deprecation Warning on OSX 10.11


juandagilc:
When I run on OS X 10.11.1 `sd.play(x, fs)` I get the next warning. Does it have to do only with sounddevice? Is there a way to remove it?

`WARNING:  140: This application, or a library it uses, is using the deprecated Carbon Component Manager for hosting Audio Units. Support for this will be removed in a future release. Also, this makes the host incompatible with version 3 audio units. Please transition to the API's in AudioComponent.h.`

mgeier:
This message is caused by the underlying PortAudio library.

Here's an issue and a mailing list discussion about the topic:
https://www.assembla.com/spaces/portaudio/tickets/243-portaudio-support-for-os-x-10-11-el-capitan/details#
https://lists.columbia.edu/pipermail/portaudio/2015-October/000092.html

I think we'll just have to wait for a new PortAudio version, until then you'll have to ignore this message.

juandagilc:
Thank's a lot!

mgeier:
You should leave this open until it is fixed.

An alternative approach would be #11, what do you think about that?

juandagilc:
I think it would be nice to remove those messages though I am not familiar with that kind of implementation.

On OSX El Capitan I could not compile PortAudio following their instructions (I could in Yosemite) and I get the errors as you showed in the links above. So as you say, the appropriate solution should come from PortAudio. I installed it using `brew install portaudio`, and so far sounddevice seems to work as I expect/need.

mgeier:
I made an attempt at silencing PortAudio in the branch "redirect-stdout", see #12.

Can you please try if this works in your case?

mgeier:
Yes, those are the right steps.

However, you should probably change the installation command (as suggested in the ["Contributing" section of the docs](http://python-sounddevice.readthedocs.org/#contributing)) to:

    python setup.py develop

... because this way you can change stuff in the code and don't need to re-install the module all the time.
You'll have to close and re-start Python, though, for the changes to take effect.

If you are not sure if the correct code is used, you should try to create an obvious error in `sounddevice.py` and check if this error is raised when you run your test script. Just to be sure ...

Since you still see the warning, maybe it is sent to `stdout` instead of `stderr` ...

Can you please try to replace the *two* occurrences of `__stderrp` with `__stdoutp` and run it again?

juandagilc:
OK, so I reinstalled as explained and the intentional error is raised!

[...]

PR #444, opened 2022-11-17
Redirect stderr with os.dup2() instead of CFFI calls

### LINKED ISSUE #90: Segmentation fault (core dumped) on Alpine Linux
I can't make that running with Alpine Linux.

Here is the dockerfile to reproduce it:
```
FROM alpine:3.6

# Setup base
RUN apk add --no-cache jq bash portaudio \
        python3 python3-dev \
        gcc libgcc \
        musl-dev musl \
        libffi-dev libffi \
    && pip3 install --no-cache --upgrade setuptools \
    && pip3 install --no-cache --upgrade sounddevice

CMD ["python3", "-m", "sounddevice"]
```

mgeier:
I'm not quite sure yet if we should do this, this has to be tested thoroughly.

With a bit of luck, this might fix #90.

However, it might also have bad side effects ...

### DIFF
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -2823,12 +2823,12 @@ def _initialize():
     is automatically called with the ``import sounddevice`` statement.
 
     """
-    devnull = None
-    old_stderr = 2  # To appease Pyright
+    old_stderr = None
     try:
         old_stderr = _os.dup(2)
         devnull = _os.open(_os.devnull, _os.O_WRONLY)
         _os.dup2(devnull, 2)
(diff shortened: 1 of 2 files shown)


mgeier:
I already mentioned this 7 years ago: https://github.com/spatialaudio/python-sounddevice/issues/10#issuecomment-158344854 and https://github.com/spatialaudio/python-sounddevice/issues/11#issuecomment-155836787

I don't quite remember the reason why I didn't go with it at the time.

Issue #72, opened 2017-02-01
query_devices() name decoding problem (MME, DirectSound)


taktot:
query_devices() crashes:

```
    File "C:\Program Files\Anaconda3\lib\site-packages\sounddevice.py", line 712, in query_devices
    _lib.Pa_HostApiTypeIdToHostApiIndex(_lib.paMME)) else 'utf-8'),
   UnicodeDecodeError: 'mbcs' codec can't decode bytes in position 0--1: No mapping for the Unicode   character exists in the target code page.
```

Changing the `"mbcs"` at line 710 to `"utf-8"` fixed the problem in my environment.

OS: Windows 7 SP1 and Windows 10 (Japanese)
Python 3.5
`sounddevice` ver 0.3.6 (wheel from http://www.lfd.uci.edu/~gohlke/pythonlibs/)

mgeier:
Can you please provide a screenshot of the output of `query_devices()` after your fix?

Does the same error happen if you run the following in a terminal?

    python3 -m sounddevice

I don't have a Windows system myself, so you'll have to help me out a bit.
I changed MME to use `'mbcs'` because of what @raecke said in https://github.com/spatialaudio/python-sounddevice/issues/30#issuecomment-236717086.
So are some MME devices using `'mbcs'` and others `'utf-8'`?

taktot:
Thank you for looking into this.

Here is the screen shot of query_devices() call.
![screenshot-querydevices](https://cloud.githubusercontent.com/assets/25470235/22573734/b002656e-e9ed-11e6-9693-bfd011505a1c.png)

And this is the result of a run from the console. The same Python error occurred.
![run-from-console](https://cloud.githubusercontent.com/assets/25470235/22573756/cd5302ea-e9ed-11e6-90e7-5b6832547473.png)

As I understand, international customization of Windows used locally devised encoding scheme (in 1980s) For instance, in Japan, Microsoft used a proprietary 2-bytes coding scheme called shift-jis.  Later, Microsoft adopted unicode as their internal string representation with tiny encoding translation layers elsewhere  to/from legacy coding scheme to talk to old applications and drivers that only speak legacy one.
So, my guess is that there are some audio driver or some old portaudio implementation that is still speaking old local encoding strings.  Unfortunately, I don't know how to identify the encoding in use from OS version etc.

raecke:
`python -m sounddevice` works with my computer without problems. My windows versionis Windows7.

```python
[...]

PR #491, opened 2023-09-07
fix decode error

### LINKED ISSUE #490: Error when ASIO devices use special encoding
The code says `All other host APIs use 'utf-8' anyway.`
This is wrong, for example on my PC there is an ASIO device that uses GBK encoding.
`b'DSD \xd7\xaa\xc2\xeb\xc6\xf7 (DoP/Native)'`
https://github.com/spatialaudio/python-sounddevice/blob/791c09db3b187257e53f79725c13c1efdb4ff458/sounddevice.py#L572-L585

btw, in my test, the `raise` at line 585 doesn't output anything, and I'm not quite sure why.
image

shoucandanghehe:
I tried to introduce the locale package to get the default encoding of the user's system.
And _I think_ there shouldn't be any encoding types other than those two.
But I can't be sure, so I just output bytes. that way there won't be any more errors caused by decoding failures.

close #490

### DIFF
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -51,12 +51,13 @@ Online documentation:
 __version__ = '0.4.6'
 
 import atexit as _atexit
+import locale as _locale
 import os as _os
 import platform as _platform
 import sys as _sys
 from ctypes.util import find_library as _find_library
-from _sounddevice import ffi as _ffi
 
+from _sounddevice import ffi as _ffi
 
 try:
     for _libname in (
@@ -570,19 +571,19 @@ def query_devices(device=None, kind=None):
     assert info.structVersion == 2
     name_bytes = _ffi.string(info.name)
     try:
-        # We don't know beforehand if DirectSound and MME device names use
-        # 'utf-8' or 'mbcs' encoding.  Let's try 'utf-8' first, because it more
-        # likely raises an exception on 'mbcs' data than vice versa, see also
-        # https://github.com/spatialaudio/python-sounddevice/issues/72.
-        # All other host APIs use 'utf-8' anyway.
+        # I tried to introduce the locale package to get the
+        # default encoding of the user's system.
+        # And I think there shouldn't be any encoding types other than those two.
+        # But I can't be sure, so I just output bytes.
+        # that way there won't be any more errors caused by decoding failures.
+        # see also 👇
+        # https://github.com/spatialaudio/python-sounddevice/issues/490
         name = name_bytes.decode('utf-8')
     except UnicodeDecodeError:
-        if info.hostApi in (
-                _lib.Pa_HostApiTypeIdToHostApiIndex(_lib.paDirectSound),
-                _lib.Pa_HostApiTypeIdToHostApiIndex(_lib.paMME)):
-            name = name_bytes.decode('mbcs')
-        else:
-            raise
+        try:
+            name = name_bytes.decode(_locale.getpreferredencoding())
+        except UnicodeDecodeError:
+            name = repr(name_bytes)[2:-1]
     device_dict = {
         'name': name,
         'index': device,
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -583,7 +583,7 @@ def query_devices(device=None, kind=None):
         try:
             name = name_bytes.decode(_locale.getpreferredencoding())
         except UnicodeDecodeError:
-            name = repr(name_bytes)[2:-1]
+            name = name_bytes.decode('ascii', errors='backslashreplace')
     device_dict = {
         'name': name,
         'index': device,


mgeier:
Thanks for this PR!

Why are you removing the `mbcs` stuff? Is it less correct than your implementation?

And I thought this only affects ASIO? Why are you applying it to all host APIs?

sounddevice.py
```diff
@@ -570,19 +571,19 @@ def query_devices(device=None, kind=None):
...
-        if info.hostApi in (
-                _lib.Pa_HostApiTypeIdToHostApiIndex(_lib.paDirectSound),
-                _lib.Pa_HostApiTypeIdToHostApiIndex(_lib.paMME)):
-            name = name_bytes.decode('mbcs')
-        else:
-            raise
+        try:
+            name = name_bytes.decode(_locale.getpreferredencoding())
+        except UnicodeDecodeError:
+            name = repr(name_bytes)[2:-1]
```
mgeier, line 586: What about using this instead?

```suggestion
            name = name_bytes.decode('ascii', errors='backslashreplace')
```

The result should be the same, but this would be more descriptive, I think.

mgeier: But maybe we don't need this fallback anyway.

I think I would prefer to get another error report (just like yours) if it turns out that some system uses yet another unexpected encoding.

shoucandanghehe:
> Why are you removing the `mbcs` stuff? Is it less correct than your implementation?

As far as I know, the default encoding for some regions of Windows is UTF-16, and mbcs can't decode it, so I tried to get the default encoding for the user's system and use it.
> And I thought this only affects ASIO? Why are you applying it to all host APIs?

I don't think judging the host API is necessary, if an error occurs decoding in UTF-8, then just try to decode it using another encoding.
If you think it will definitely be UTF-8 encoded, then it won't even get to that if.
If it's not UTF-8, the worst that can happen is that an exception is thrown again, (if we don't use `repr`) but if it's the default encoding, we're good to go.

sounddevice.py
shoucandanghehe: Maybe we can keep this fallback, but use `print(e)` to output the error so the user can continue to use it or come and submit the error.

mgeier:
> As far as I know, the default encoding for some regions of Windows is UTF-16

This may well be, but do you know for sure whether each single host API actually uses the default encoding?
Did you test this?

> I don't think judging the host API is necessary

It may not be, but I don't know that for sure.

I only know that this was tested on multiple systems in #72.

And AFAIU, you only saw the problem with ASIO and not with any other host API, right?

The problem is that I cannot test any of this myself and I have to rely on reports by other people.

sounddevice.py
mgeier: I'm strongly against using `print()`.

shoucandanghehe:
> This may well be, but do you know for sure whether each single host API actually uses the default encoding? Did you test this?

I certainly can't test all versions of Windows, so I added the fallback just so I can continue to use it if the name decoding fails.

> And AFAIU, you only saw the problem with ASIO and not with any other host API, right?

Yes, but I don't think it's going to get any worse, for the reasons I stated above, utf-8 is still the first-order decoding scheme.

mgeier:
> > This may well be, but do you know for sure whether each single host API actually uses the default encoding? Did you test this?
> 
> I certainly can't test all versions of Windows, so I added the fallback just so I can continue to use it if the name decoding fails.

Yeah, but maybe this leads to worse behavior (unreadable device names) in some cases that did work before. We don't know that, but I would like to not risk that.

> PS. I think it's very frustrating to have an entire program unavailable because the name of some mysterious device was decoded incorrectly.

I agree. That's why it would be great to fix the error with ASIO that you actually witnessed.

However, it would also be frustrating if the program does not crash but instead some device names contain gibberish.
Using the suggested fallback too often might lead to that problem.

In either case (crash or gibberish), I have to rely on user feedback (like yours) to be able to fix the problem.
I just think that a crash is a better incentive for people to report a problem.

And I'm open to fix every single problem that comes up. I just want to be cautious not to re-introduce errors on systems that have been working correctly before.

> > And AFAIU, you only saw the problem with ASIO and not with any other host API, right?
> 
> Yes, but I don't think it's going to get any worse, [...]

That's the problem, I don't know that. Maybe it gets worse, maybe it doesn't. That's why I would like to keep the behavior that has actually be tested and confirmed to work correctly and only fix the behavior that has actually been observed to be broken.

mgeier:
I have created an alternative to this PR: #512.

@shoucandanghehe Can you please check if that also fixes your problem?

shoucandanghehe:
> @shoucandanghehe Can you please check if that also fixes your problem?

Actually, I don't know what happened to my driver, anyway that device disappeared😇

mgeier:
I have merged #512 instead.

Issue #496, opened 2023-10-13
Importing sounddevice interrupts other audio streams on Windows when exclusive mode is enabled


arkrow:
As the title mentions, importing sounddevice interrupts any currently playing audio streams momentarily on Windows when exclusive mode is enabled on the sound device. This issue seems to originate from PortAudio's initialization logic, however, I can't seem to pinpoint the problematic code there, and there isn't a similar issue opened there either.

To replicate the issue:
1. On Windows 10/11, enable exclusive mode on the default sound device.
2. Run any audio stream in the background (it doesn't seem necessary for the stream to be using exclusive mode)
3. `import sounddevice` (can be run in the python interactive interpreter; the current audio stream will be interrupted as a result)

This issue does not occur if exclusive mode is disabled in the sound device. I'd assume this issue can be reproduced on Windows 7 as well.

mgeier:
> This issue seems to originate from PortAudio's initialization logic, however, I can't seem to pinpoint the problematic code there, and there isn't a similar issue opened there either.

Then you should open one!

This doesn't seem like anything that can be solved from the Python side.

To be sure, you could replace your step 3 with some other software that uses PortAudio, for example Audacity, and listen if it behaves the same.

arkrow:
> To be sure, you could replace your step 3 with some other software that uses PortAudio, for example Audacity, and listen if it behaves the same.

Great call. I've tested the same scenario with Audacity and it doesn't exhibit this issue. In fact, using Audacity's PortAudio binary (instead of the ones linked in `_sounddevice_data`) solves the issue. From what I can gather, Audacity's PortAudio binary is based on the 19.7 release, same as here, confirmed by comparing `.Pa_GetVersion()` for both libraries (though the revision is unknown for Audacity's binary). Regardless, the discrepancy in behavior is unexpected.

mgeier:
Hmm, that's interesting.

I don't know exactly how the DLL from Audacity is built, probably with Conan. I found https://github.com/audacity/audacity/blob/52708c770ffcd98cb86527cd7756413c4179f7fe/conan/conanfile.py#L119-L121, which suggests that WDM-KS is disabled. Maybe that causes the problem? It could be something completely different, though.

[...]

Issue #130, opened 2018-04-04
Allow a way to specify the portaudio location


peircej:
It would be nice to be able to specify the location of the portaudio lib for people that want to use a specific location and/or version. At present the code just uses whatever ctypes find_library discovers or reverts to the provided `_sounddevice_data` folder.

Downsides:

1. it's often unnecessary to provide portaudio (at least on *nix systems)
2. this isn't the location that the rest of the OS expects things. That causes problems for packaging the lib into an app. For example to wrap sounddevice up into an application bundle on Mac (e.g. py2app) the dylib files really should go into the Contents/Frameworks folder. If the app is going to be code-signed to verify it was from a trusted developer then the dylibs MUST go into the frameworks folder

I wonder if we could have something like a check for an environment variable first, like os.environ['SD_PORTAUDIO'] ?

j9ac9k:
I just created a PR for PyInstaller to bundle `libportaudio.dylib` as a binary and place it in the correct location within a `.app` bundle.  You can follow along w/ the PR here: 

https://github.com/pyinstaller/pyinstaller/pull/4498

mgeier:
Thanks @j9ac9k!

j9ac9k:
Thanks for making this library, you helped me regain my sanity with trying to replicate the `sounddevice.OutputStream` functionality with the Qt framework using `QAudioOutput` (after weeks of trying I gave up, and stumbled across sounddevice).

Also in case you're curious, with this PR, `libportaudio.dylib` is placed in a part of the `.app` bundle where the bundle can be codesigned and notarized.

papr:
Hi!

I am using PyInstaller to bundle a `sounddevice` application on macOS, Windows, and Ubuntu. It works great on Windows and macOS but fails **on Linux** because `ctypes.util.find_library` is not able to find the library at runtime.

#### Sanity Checks
I have made sure that
1. `libportaudio2` is installed at packaging time
2. `libportaudio.so.2` is packaged correctly in the application
3. `LD_LIBRARY_PATH` includes the path to the `libportaudio.so.2` directory at runtime

#### Experiment

I have tested the following within the frozen application:
| Search term |  `ctypes.util.find_library` | `ctypes.CDLL` | `_sounddevice.ffi` |
|-|-|-|-|
| portaudio | ❌ | ❌ | ❌ |
| libportaudio | ❌ | ❌ | ❌ |
| libportaudio.so | ❌ | ❌ | ❌ |
[...]

PR #498, opened 2023-10-25
Add an environ flag to load an alternative PortAudio Windows binary compiled without ASIO

arkrow:
This PR provides a very simple and minimal workaround for loading an alternative PortAudio Windows binary compiled without ASIO, to address issue #496. Since this issue only affects Windows, as far as I know, no changes have been made to the other platforms.

An alternative PortAudio binary will need to be added to the `_sounddevice_data` repository (e.g. named `libportaudio64bit-noasio.dll`). It will then be bundled like the other binaries in the built package and can then be used instead of the default using the `os.environ["SD_WIN_DISABLE_ASIO"]` flag.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,6 +23,7 @@ if system == 'Darwin':
     libname = 'libportaudio.dylib'
 elif system == 'Windows':
     libname = 'libportaudio' + architecture0 + '.dll'
+    libname_noasio = 'libportaudio' + architecture0 + '-noasio.dll'
 else:
     libname = None
 
@@ -30,6 +31,9 @@ if libname and os.path.isdir('_sounddevice_data/portaudio-binaries'):
     packages = ['_sounddevice_data']
     package_data = {'_sounddevice_data': ['portaudio-binaries/' + libname,
                                           'portaudio-binaries/README.md']}
+    if system == 'Windows':
+        package_data['_sounddevice_data'].append(
+            'portaudio-binaries/' + libname_noasio)
     zip_safe = False
 else:
     packages = None
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -74,7 +74,10 @@ except OSError:
     if _platform.system() == 'Darwin':
         _libname = 'libportaudio.dylib'
     elif _platform.system() == 'Windows':
-        _libname = 'libportaudio' + _platform.architecture()[0] + '.dll'
+        if "SD_WIN_DISABLE_ASIO" in _os.environ:
+            _libname = 'libportaudio' + _platform.architecture()[0] + '-noasio.dll'
+        else:
+            _libname = 'libportaudio' + _platform.architecture()[0] + '.dll'
     else:
         raise
     import _sounddevice_data


mgeier:
Thanks for this PR!

I was hoping for a more general solution for choosing the DLL/dylib/so, similar to how it was discussed in #130.

But before even going down the environment variable path, I would like to know how `ctypes.util.find_library()` would find a user-provided DLL without having to use an env var.

I know that this works on Linux and I assume that it works on macOS as well and it would be good to know how to make this work on Windows as well. Maybe that would already be enough for selecting a custom DLL?

arkrow:
> I was hoping for a more general solution for choosing the DLL/dylib/so, similar to how it was discussed in #130.

I think the most promising approach would be building libportaudio as part of the wheel build process, as mentioned in #130, and accepting ASIO as a flag in package extras as I mentioned in #496. If that can work on all systems without additional prerequisites, that would be the ideal solution. If I can get the solution up and running, I'll open a new PR.

Alternatively, as part of the suggested solutions in #130, I've opened a PR (#499) that loads a user-specified PortAudio binary during runtime, which is more general than this PR and has been requested by other users.

> But before even going down the environment variable path, I would like to know how `ctypes.util.find_library()` would find a user-provided DLL without having to use an env var.

That would be difficult to do dynamically (without hard-coding) unless the values is retrieved from somewhere, which is why `os.environ` is a viable and often used approach. `os.environ` is also not permanent and its scope extends to the current python process only, avoiding any problematic PATH changes, so it should be safe to use on all platforms.

mgeier:
> That would be difficult to do dynamically (without hard-coding)

Do you mean hard-coding the absolute path to the DLL?
What about just hard-coding a file name?

According to https://docs.python.org/3/library/ctypes.html#finding-shared-libraries:

> On Windows, [find_library()](https://docs.python.org/3/library/ctypes.html#ctypes.util.find_library) searches along the system search path, and returns the full pathname, but since there is no predefined naming scheme a call like `find_library("c")` will fail and return `None`.

So if you put a correctly named DLL somewhere into the *system search path* (whatever that means exactly), it should be found, right?
Do you know what the *system search path* is?

In case this hasn't been obvious already: I'm not a Windows user, so you can assume zero knowledge from my side.

mgeier:
See my comment https://github.com/spatialaudio/python-sounddevice/issues/496#issuecomment-1902706025

This doesn't provide an environment variable, but it involves downloading the appropriate DLL, renaming it and moving it into a specific folder.

I think that would still be easy enough for the (few?) people who need ASIO, what do you think?

mgeier:
In #539, I'll switch to DLLs without ASIO by default. This also adds documentation for how to download and use DLLs with ASIO, if needed.

I think this makes this PR obsolete, right?

arkrow:
Since the default behavior is without ASIO, this PR is now obsolete. Though it's inverse can be considered if both binaries are to be shipped together and enabled conditionally; that might be a good middle-ground for developers depending on ASIO support with this library. That said, the PR will be closed for now.

PR #499, opened 2023-10-30
Load user-provided PortAudio name/path from environ's SD_PORTAUDIO

arkrow:
Implements the necessary functionality suggested in #130


mgeier:
I have documented how to load a custom PortAudio library: #518

@arkrow Do you think your PR is still needed, since users can already use custom DLLs etc.?

arkrow:
> I have documented how to load a custom PortAudio library: #518
> 
> @arkrow Do you think your PR is still needed, since users can already use custom DLLs etc.?

One advantage this implementation has is that it solves potential PATH shadowing issues (e.g. another directory with portaudio is in PATH and has higher precedence/order, so it is loaded instead of the custom portaudio binary that a developer shipped with their package/software and intended to use).

I think it'd be good to continue the discussion in #130. I can pick up this PR again in the future if you are open to this implementation and other members of the community find it useful for their use case.

mgeier:
In the meantime, it is ...

* documented how to load an arbitrary DLL: https://python-sounddevice.readthedocs.io/en/0.5.0/installation.html#custom-portaudio-library
* possible to select the bundled ASIO-enabled DLL via environment variable: #556

Do we still need this PR?

arkrow:
Hey @mgeier,

Thanks for following up with the documentation on the other PR! I was occupied lately and unfortunately couldn't continue till now.

This PR is no longer necessary, since there is no current use case or demand for loading custom PortAudio binaries dynamically just before import time. The windows ASIO use case is covered on the other PR.

I'll be closing the PR for now. Since the logic implemented here is quite simple, it can be picked up again if needed by other contributors.

Issue #507, opened 2023-12-01
Enhancement Request: Integrate Playback Status Check into Simple `play`/`stop` API in `sounddevice`


elias-jhsph:
## Summary
This issue proposes integrating a feature for checking the playback status into the simpler `play`/`stop` API of the `sounddevice` module. While the module does offer more robust methods for handling audio playback, having this functionality readily available in the simpler API would greatly enhance usability for basic use cases.

## Background
Currently, in my application, I use the `sounddevice` module's simple API for playing audio files. However, to check if the playback has finished, I have to resort to accessing a private member (`_last_callback`), as shown below:

```python
import sounddevice as sd
import soundfile as sf

# ... middle of function doing other things
for loop in range(loops):
    data, fs = sf.read(file_path)
    sd.play(data, fs)
    while not sd._last_callback.event.is_set():
        if stop_event.is_set() or (added_stop_event and added_stop_event.is_set()):
            sd.stop()
            break
        if added_stop_event:
            added_stop_event.wait(timeout=.1)
        else:
            stop_event.wait(timeout=.1)
```

## Recognizing the More Robust API
I am aware that sounddevice offers a more robust API that could potentially handle these scenarios more effectively. However, for simpler applications or for users who are not deeply versed in audio programming, the complexity of the robust API can be a barrier.

## Proposal
I propose enhancing the simpler play/stop API by adding a method or property to check the playback status. This could be a method like is_playing() or a property like playback_status. This enhancement would not only make the code more readable and maintainable but also provide a more accessible way for users to handle basic audio playback tasks.

## Example of Proposed Solution
With the proposed enhancement, the implementation would look like this:

```python
import sounddevice as sd
import soundfile as sf

# ... middle of function doing other things
for loop in range(loops):
    data, fs = sf.read(file_path)
    sd.play(data, fs)
    while sd.is_playing():  # Proposed method
        if stop_event.is_set() or (added_stop_event and added_stop_event.is_set()):
            sd.stop()
            break
        if added_stop_event:
            added_stop_event.wait(timeout=.1)
        else:
            stop_event.wait(timeout=.1)
```

## Conclusion
Integrating a playback status check into the simple play/stop API would not only cater to users seeking simplicity but also make the sounddevice module more versatile and accessible to a broader range of users, from beginners to advanced programmers.

mgeier:
What about this:

```python
while sd.get_stream().active:
    # ...
```

[...]

PR #509, opened 2023-12-03
Implement Timeout Feature for sounddevice.wait() Function

elias-jhsph:
The purpose of this pull request is to introduce a new feature to the `sounddevice` module's `wait()` function. This update adds an optional `timeout` parameter, allowing users to specify a maximum waiting time for audio playback or recording to finish.

Key Highlights:
- Introduces a `timeout` parameter to the `wait()` function, enabling users to set a maximum wait time.
- Enhances the usability of the `sounddevice` module, particularly in scenarios where blocking the thread indefinitely is not viable.
- Ensures backward compatibility by setting the default value of `timeout` to `None`, which retains the original behavior of the `wait()` function.
- Includes updated documentation and docstrings explaining the new feature and how to use it.
- The changes have been thoroughly tested to ensure reliability and consistency with the existing functionality of the module.

This resolves issue #507 

Thank you for considering this contribution to the `sounddevice` module. I look forward to your feedback and the opportunity to improve this essential audio handling tool in Python.

### DIFF
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -403,7 +403,7 @@ def wait(ignore_errors=True, timeout=None):
           exception raised.
         - False if timeout is specified and timeout elapses before
           playback/recording finished and no exception raised.
-        
+
     See Also
     --------
     get_status


mgeier:
Thanks for this PR!

I'm not sure if it is good to have two falsey return values (`None` and `False`) for two different outcomes (success and timeout).

Could you please show a few hypothetical uses that illustrate how to deal with those return values?

mgeier:
> But if you want some examples on that I can happily provide them.

Yes, please.

Currently, I'm not interested in thinking about alternatives, I would first like to see how your current suggestion looks when it is used.

elias-jhsph:
Old behavior is preserved but here is some examples of how timeout could be used

## Interactive Audio Guide for a Museum Exhibit
Imagine an interactive audio guide system in a museum where visitors can trigger audio descriptions of various exhibits. Since the length of these audio clips can vary and visitors might interact with multiple exhibits simultaneously, the system needs to check for new inputs while playing the current audio clip.

```python
import sounddevice as sd

def load_exhibit_audio(exhibit_id):
    # Load the audio file for the given exhibit
    return ...  # Replace with actual audio loading logic

def check_for_new_interaction():
    # Check if a visitor has interacted with a new exhibit
    return ... # Replace with actual interaction check logic

# Initial setup
current_exhibit_id = "exhibit_1"
audio_clip, samplerate = load_exhibit_audio(current_exhibit_id)
sd.play(audio_clip, samplerate)

poll_speed = 2  # Interval for each check (seconds)

# Main loop
while True:
    new_interaction = check_for_new_interaction()
    if new_interaction and new_interaction != current_exhibit_id:
        sd.stop()
        print(f"New interaction detected: {new_interaction}")
        current_exhibit_id = new_interaction
        audio_clip, samplerate = load_exhibit_audio(new_interaction)
        sd.play(audio_clip, samplerate)

    finished = sd.wait(timeout=poll_speed)
    if finished != False:
        if finished != True:
            print("Restarting...")
            break
        else:
            print("Waiting for next interaction...")
            time.sleep(poll_speed)

print("Audio guide session ended.")
```

## Environmental Interaction with Audio Playback and Monitoring
A research setup plays an audio clip (like a bird song) to study its influence on animal behavior. Simultaneously, a separate sensor system monitors the environment for any responses or changes.

```python
import sounddevice as sd
import time

# Prepare an audio clip (e.g., bird song)
bird_song, samplerate = ...  # Load or generate bird song audio

def check_environmental_response():
    return # Logic to receive data from the sensor system

# Start playing the bird song
sd.play(bird_song, samplerate)

monitoring_duration = 30  # Total duration for monitoring (seconds)
timeout_duration = 5      # Interval for each check (seconds)
start_time = time.time()

finished = False
while finished:  
    response_detected = check_environmental_response()
    if response_detected:
        print("Environmental response detected!")
    else:
        print("No response detected yet.")
    finished = sd.wait(timeout=timeout_duration)
    if time.time() - start_time > monitoring_duration:
    	sd.stop()
    	break
    	
if finished == False:
    print("Bird song cut off because it was longer than monitor duration")
elif finished != True:
    print("Bird song cut off due to issue during playback")
else:
    print("Bird song finished")
```

mgeier:
Thanks for these examples!
I think they will be very useful for evaluating the API.

There is a little bug, but it isn't really relevant for us: I think it should be `while not finished`.

And there are a few instances where `pycodestyle` complains:

    E712 comparison to False should be 'if cond is not False:' or 'if cond:'

Using `==` and `!=` with `True` and `False` is un-Pythonic.

Apart from that, I would be interested how the `CallbackFlags`-part of the result could be handled.
IMHO it is a cardinal sin to ignore buffer over-/underruns.
And in case there is no timeout, I think it would be appropriate to check for that.

mgeier:
Don't worry, it doesn't really have to be runnable. It's only about the line with the `sd.wait(timeout=...)` call and the following handling of its return value. 

> Do you need anything more from me

Yes, sorry that I wasn't clear ... it would be great if you could provide variations of your examples that handle the `CallbackFlags` thing.
Because this will make the handling of the return value even more complicated.

Also, it would be nice if you could fix the thing that `pycodestyle` complains about, because things like `finished == False` look really wonky.

And then I would be interested in your opinion whether that usage looks nice, or if see any potential for improvement.

mgeier:
> After self.stream.close are callback flags reset?

No, they should stay available to be queried by the user until the next stream is created.

> What action make sense if some callback flags are set for this level of the api?

I think the most important action is somehow reporting or logging the problem.
I guess the easiest to do is something like this:

```python
if flags:
    print(flags)
```

But that's the problem with returning `True`, which complicates this check.

For this PR it doesn't really matter what the user does with this information, as long as they can find out whether some flags are set or not and react accordingly.

> Will sounddevice raise exceptions while the wait is running

No, I don't think so.

mgeier:
I also wanted to comment on something you said earlier:

> It also has the added benefit of matching return behavior of the standard threading library “wait” in this scenario.

The behavior of different `timeout` parameters in the standard library is surprisingly inconsistent.
Sometimes `True` or `False` are returned, but there is also https://docs.python.org/3/library/asyncio-task.html#asyncio.wait_for and https://docs.python.org/3/library/queue.html#queue.Queue.get, which raise different exceptions on timeout.

Maybe this would also make sense here?

It might look something like this:

```python
while True:
    if ...:
        sd.play(...)

    try:
        status = sd.wait(timeout=poll_speed)
    except sd.TimeoutExpired:
        print("Waiting for next interaction...")
    else:
        if status:
            print(status)
        print("Restarting...")
        break
```

mgeier:
@elias-jhsph Any news on this? What do you think about my comments?

elias-jhsph:
I was busy with work for my job but I am picking this back up this week
________________________________

mgeier:
@elias-jhsph Did you have a chance to look at this in the meantime?

Issue #513, opened 2023-12-22
Sounddevice + big dict + OOP = Segfault / Bus Error


theredled:
I'm having a very strange problem, which I managed to reduce as much as possible like this:

```python
import sounddevice
import time

class SamplerBox:
    def __init__(self):
        self.samples = {}

    def audio_callback(self, outdata, frame_count, time_info, status):
        print('ac')

    def init(self):
        self.connect_audio_output()
        self.load_samples()
        time.sleep(20)

    def connect_audio_output(self):
        try:
            sd = sounddevice.OutputStream(callback=self.audio_callback)
            sd.start()
            print('Opened audio device')
        except:
            print('Invalid audio device')
            exit(1)

    def load_samples(self):
        for midinote in range(128):
            for velocity in range(128):
                self.samples[midinote, velocity] = Sound()


class Sound:
    def __init__(self):
        pass


sb = SamplerBox()
sb.init()

```

As soon as I create that big `self.samples` dict, and only create a new audio stream with an empty callback, I get "Bus Error 10" with Python 3.11. 

With Python 3.9 I get "Illegal instruction 4"

In my original script (reduced here) I got "Segmentation Fault 11"

I'm running Homebrew Python 3.11 on MacOS 10.15.7.

Worst than that, ***written in a procedural way, it runs perfectly*** :

```python
import sounddevice
import time

samples = {}


class Sound:
    def __init__(self):
        pass

def audio_callback(self, outdata, frame_count, time_info, status):
    print('ac')


try:
    sd = sounddevice.OutputStream(callback=audio_callback)
    sd.start()
    print('Opened audio device')
except:
    print('Invalid audio device')
    exit(1)

for midinote in range(128):
    for velocity in range(128):
        samples[midinote, velocity] = Sound()

time.sleep(20)
```

Any idea?

mgeier:
This indeed looks quite curious, and I can reproduce it on Linux, however only if I increase the range of the `for` loops (I got a segmentation fault with `1280`, but maybe it also crashes with lower values).

I don't really know what's happening and I think it *shouldn't* happen ... but I think I know how to avoid the problem:

In your first example, `sd` is a local variable that goes out of scope at the end of `connect_audio_output()`, and I guess at some point the Python interpreter destroys the stream object while the callback is still running or something.

You should assign the stream object as an instance variable of the `SamplerBox` object, something like this:

```python
self.sd = sounddevice.OutputStream(...)
```

[...]

PR #514, opened 2023-12-22
Raise an assertion if an active stream is destroyed

mgeier:
I'm not quite sure if that's the proper thing to do, but I hope it's not too disruptive compared to the previous behavior.

This does *not* avoid the problem in #513, but it hopefully makes it easier to diagnose the problem.

### DIFF
--- a/sounddevice.py
+++ b/sounddevice.py
@@ -1106,6 +1106,9 @@ class _StreamBase:
         self.stop()
         self.close()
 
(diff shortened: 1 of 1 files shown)


mgeier:
Can somebody please confirm that this improves their situation (an how)?
Otherwise I will close this without merging.

mgeier:
I'm closing this due to lack of response.

PR #548, opened 2024-08-11
Change DeviceList repr to use correct indices

rmccampbell:
The index from enumerate won't match the actual device index if the list is filtered/reordered for whatever reason (i.e. to restrict to inputs/outputs or a specific API). This also means the default mark would be wrong. This fix uses the index in the dictionary instead.


mgeier:
Thanks!

PR #556, opened 2024-09-04
Add an environ flag to load an ASIO PortAudio binary

arkrow:
This PR adds the option to load an alternative PortAudio binary with ASIO support, enabled using an environ flag.

Additional PortAudio binaries compiled with ASIO will need to be maintained in the `_sounddevice_data` repository (e.g. named `libportaudio64bit-asio.dll`). They will then be bundled like default binary in the built package and can then be used by using the os.environ["SD_ENABLE_ASIO"] flag.

This PR is based on, and is the inverse of #498, since the default is now ASIO-disabled.

Minimal example:

```py
import os

# Set environ key anywhere before importing sounddevice. Value is not important.
os.environ["SD_ENABLE_ASIO"] = "1"

# Since "SD_ENABLE_ASIO" is set, sounddevice will load the ASIO-enabled binary
import sounddevice as sd

print(sd.query_hostapis())
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,6 +23,7 @@ if system == 'Darwin':
     libname = 'libportaudio.dylib'
 elif system == 'Windows':
     libname = 'libportaudio' + architecture0 + '.dll'
+    libname_asio = 'libportaudio' + architecture0 + '-asio.dll'
 else:
     libname = None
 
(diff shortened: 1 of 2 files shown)


mgeier:
Thanks @arkrow for this PR and thanks @rob-broadley for testing!

This will not work in the `conda` package, which is no problem (at least not ours), but we should probably still mention that in the docs? (The docs already mention that the `conda-forge` package doesn't have ASIO support, but maybe it doesn't hurt to also mention it in the context of the environment variable)

It will also not work if a custom `portaudio.dll` is present in the `%PATH%`. I think this is also fine, but should maybe also be mentioned in the docs.

@arkrow Would you like to add some documentation?

mgeier:
Thanks again! I will add the docs in a separate PR.

mgeier:
I have added some documentation in #558.

PR #560, opened 2024-10-07
pyproject: Use pyproject.toml for building wheels

abelgladstone:
Remove setup.py and move to pyproject.toml


mgeier:
Thanks for this PR!

I like this approach in general, but it seems that the "cross-platform package creation" doesn't work anymore.
This means that the CI doesn't create platform-specific wheels anymore:
https://github.com/spatialaudio/python-sounddevice/blob/2748f7049a36e7ada23de2f39c4bb2e5038c4796/.github/workflows/publish.yml#L18-L19

abelgladstone:
@mgeier thank you for the reply. I will try make an effort to get the platform specific builds to work as well. Will get back to you soon.

abelgladstone:
Unfortunately my little home PC is stuck on windows. As a consequence of intercepting the filename. I had to move the version to a separate VERSION file. So that I can read it without any fuss. I hope this helps.

mgeier:
Thanks for the quick update!

I need some more time to review and test this thoroughly, but in the meantime I was wondering whether we should enable building the wheels on PRs: #561. Because otherwise we don't know whether this PR really works, right?

_custom_build/backend.py
```diff
@@ -0,0 +1,58 @@
...
+    elif current_platform == WINDOWS:
+        oses = "win32" if architecture == "32bit" else "win_amd64"
+
+    with open("VERSION", "r") as f:
+        version = f.read().strip()
+
+    return f"sounddevice-{version}-py3-none-{oses}.whl"
+
+
+def prepare_meta_for_build_wheel(config_settings=None):
```
mgeier, line 37: I guess this should be ...

```suggestion
def prepare_metadata_for_build_wheel(config_settings=None):
```

... but the whole function seems unnecessary, since it is imported already?

abelgladstone:
Agreed.

abelgladstone:
Done

mgeier:
Thanks!

The wheels are available here: https://github.com/spatialaudio/python-sounddevice/actions/runs/11293700433?pr=560

Now all of them contain all binaries, but each should only contain the binaries for their own platform.

pyproject.toml
```diff
@@ -1,3 +1,41 @@
...
+
+[tool.setuptools]
+zip-safe = false
+py-modules = ["sounddevice", "_sounddevice"]
+packages = ["_sounddevice_data"]
+
+[tool.setuptools.package-data]
+"_sounddevice_data"= [
+        "portaudio-binaries/*.dll",
+        "portaudio-binaries/*.dylib",
```
mgeier, line 36: I guess this causes all wheels to contain all binaries. This should be limited.

Also, the `sdist` now also contains the binaries, but it shouldn't.

abelgladstone:
Thanks for the review. I'll get back to you soon.

abelgladstone:
After many tries I am unable to find a way to build the project without the setup.py file in which case what is in the project is already good enough IMO.

mgeier:
Thanks a lot for your efforts anyway!

AFAIU, the problem is that `package-data` cannot be set dynamically depending on the platform, while that is possible in `setup.py` with `package_data`.

Maybe future changes to `setuptools` will make this possible, or maybe we'll find a different build tool that supports this.

If anyone has an idea, please speak up!

PR #573, opened 2024-11-27
Add codespell support (config, workflow to detect/not fix) and make it fix few typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

You barely had any typos (well done!) so feel free to ignore/close this PR -- was not much effort.

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,25 @@
+# Codespell configuration is within pyproject.toml
+---
+name: Codespell
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+permissions:
+  contents: read
+
+jobs:
+  codespell:
+    name: Check for spelling errors
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Annotate locations with typos
+        uses: codespell-project/codespell-problem-matcher@v1
+      - name: Codespell
+        uses: codespell-project/actions-codespell@v2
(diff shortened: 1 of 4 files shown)


mgeier:
Thanks for this PR!

I'm a bit torn, though. On the one hand, I like using automation to fix errors, but OTOH I'm worried about the UX regarding false positives.

There is already one false positive that has to be handled, and one case that is arguable (see https://github.com/codespell-project/codespell/issues/3521 and https://github.com/sphinx-doc/sphinx/pull/13113/files#diff-5ac55e1c4199fe776eedaf542c066d6b3ff665c930c565c860e9b4710704a8cd). This leaves just a single uncontested error.

This makes me think that the probability of false positives appearing in the future is quite high. Handling those would probably be annoying for potential contributors.

That's what I like about https://pep8speaks.org/: it makes it easy to ignore suggestions (it doesn't cause CI to fail) and it only makes suggestions for the lines that were actually changed in a given PR. I guess that is not possible with codespell, right?

yarikoptic:
it would be close to that (only check the files modified) if you start using and add codespell to pre-commit which (by default) would make it go only through modified files.  Given that overall there is not that many typos , feel welcome to close if feeling like it! no hurt feelings ;)  but as for false positives -- the rate is actually surprisingly low and it is relatively easy to deal with them if they come up.

Cheers

mgeier:
I've cherry-picked the actual spelling fixes to #580.

I think the rest is too much procedure for too little predicted gain.

But thanks anyway for your efforts!

> add codespell to pre-commit which (by default) would make it go only through modified files.

This doesn't help much in this case, since we basically only have a single file.

PR #583, opened 2025-04-30
Drop support for Python 3.7

mgeier:
It's end of life and it doesn't seem to be supported anymore by GitHub Actions.


mgeier:
I've merged #584 instead, but I guess sooner or later we'll want to get rid of Python 3.7 anyway.

PR #596, opened 2025-08-27
Move parts of setup.py to pyproject.toml

mgeier:
This moves a subset of #560, which was suggested by @abelgladstone.

I still don't know a way how to completely get rid of `setup.py`, but this partial move already has advantages (for me, it's using "editable" dependencies with `uv`).


mgeier:
This doesn't seem to work for "editable" installations. When running `uv sync`, the file `_sounddevice.py` is not generated.

mgeier:
I looked into this again, and I found out that editable installations didn't work even before this PR (see #602)!

I've updated the developer installation instructions with a work-around.

While I was at it, I also switched to the `src/` layout (which I wanted to do for a long time).

Thorough testing would be appreciated!

PR #597, opened 2025-08-30
Add explicit sample format for WASAPI

kaidokert:
Adds `explicit_sample_format`->`paWinWasapiExplicitSampleFormat` flag, which allows correctly querying a device that supports for example only 16-bit format.

Without this flag every `dtype=` is claimed to be supported by `check_input_settings` and opening the device with exclusive mode, without auto conversion will fail.


mgeier:
Thanks!

PR #628, opened 2026-04-22
Fix broken file descriptors in frozen windows apps

### LINKED ISSUE #461: Unable to use subprocess.check_output with sounddevice 0.4.6
Unable to call subprocess.check_output after importing sounddevice 0.4.6
Output error: [WinError 50] The request is not supported

System: Windows 10 and 11
Python: 3.11.2
sounddevice 0.4.6
pyinstaller 5.9.0

Easy reproduction

1. Write script
```
import subprocess
output = subprocess.check_output("ECHO before import", shell=True)
print(output)
import sounddevice
output = subprocess.check_output("ECHO after import", shell=True)
print(output)
```

2. Build using pyinstaller 5.9.0

`python -m PyInstaller --onefile aaa.py   
`

```
aaa.exe
> b'before import\r\n'
> Traceback (most recent call last):
> File "aaa.py", line 7, in <module>
> File "subprocess.py", line 466, in check_output
> File "subprocess.py", line 548, in run
> File "subprocess.py", line 892, in __init__
> File "subprocess.py", line 1378, in _get_handles
> File "subprocess.py", line 1387, in _make_inheritable
> OSError: [WinError 50] The request is not supported
```

ndahn:
Fixes #461.

### DIFF
--- a/src/sounddevice.py
+++ b/src/sounddevice.py
@@ -2925,13 +2925,17 @@ def _initialize():
 
     """
     old_stderr = None
-    try:
-        old_stderr = _os.dup(2)
-        devnull = _os.open(_os.devnull, _os.O_WRONLY)
-        _os.dup2(devnull, 2)
-        _os.close(devnull)
-    except OSError:
-        pass
+    # Duping std file descriptors in a frozen context on windows will corrupt
+    # the fd and cause issues with e.g. subprocess.check_output. This happens
+    # for example in a compiled pyinstaller exe. See #461 for details.
+    if _sys.platform != 'win32' or not getattr(_sys, 'frozen', False):
+        try:
+            old_stderr = _os.dup(2)
+            devnull = _os.open(_os.devnull, _os.O_WRONLY)
+            _os.dup2(devnull, 2)
+            _os.close(devnull)
+        except OSError:
+            pass
     try:
         _check(_lib.Pa_Initialize(), 'Error initializing PortAudio')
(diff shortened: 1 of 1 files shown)


src/sounddevice.py
```diff
@@ -2925,13 +2925,17 @@ def _initialize():
...
-        old_stderr = _os.dup(2)
-        devnull = _os.open(_os.devnull, _os.O_WRONLY)
-        _os.dup2(devnull, 2)
-        _os.close(devnull)
-    except OSError:
-        pass
+    # Duping std file descriptors in a frozen context on windows will corrupt
+    # the fd and cause issues with e.g. subprocess.check_output. This happens
+    # for example in a compiled pyinstaller exe. See #461 for details.
+    if _sys.platform != 'win32' or not getattr(_sys, 'frozen', False):
```
mgeier, line 2931: It's probably a matter of taste, but I think this would be easier to understand:

```suggestion
    if not (_sys.platform == 'win32' and getattr(_sys, 'frozen', False)):
```

Or maybe you could do something like this:

```python
    if _sys.platform == 'win32' and getattr(_sys, 'frozen', False):
        # the whole comment from above, explaining what's wrong in this case
        pass
    else:
        # the original redirection code
```

What do you think?

ndahn: Yeah, I agree that's cleaner. I'll update it once I have time to make an MRE, too.

PR #629, opened 2026-05-04
Fix callback handling and flag initialization

IzaCarlos:
## Problem
On Python 3.14, cffi changed how it passes arguments to native callbacks
and how it validates `__init__` return values. This caused two distinct errors:

1. `TypeError: __init__() should return None, not 'NoneType'`
   cffi now intercepts and validates the return of `__init__` at a lower
   level, causing it to spuriously reject the standard `None` return.

2. `TypeError: unsupported operand type(s) for &: '_CDataBase' and 'int'`
   The `status` argument now arrives as a raw `_CDataBase` object instead
   of a plain Python int, breaking bitwise operations in `_hasflag`.

Both errors surface in `_wrap_callback` and occur unpredictably during
stream callbacks, including during silence and active playback.

## Fix
Bypass `CallbackFlags.__init__` entirely in `_wrap_callback` using
`__new__`, then set `_flags` directly via `int()` cast. This avoids
the cffi `__init__` interception and the `_CDataBase` bitwise issue
in one shot.

## Changes
In `_wrap_callback`, replace:

    args = args[:-1] + (CallbackFlags(args[-1]),)
With:

    cf = CallbackFlags.__new__(CallbackFlags)
    cf._flags = int(status)

## Tested on
- Python 3.14.4, Sounddevice 0.5.5
- Extended idle periods (no audio) no longer trigger the error
- Active playback confirmed working after fix

### DIFF
--- a/src/sounddevice.py
+++ b/src/sounddevice.py
@@ -2770,12 +2770,13 @@ def _get_stream_parameters(kind, device, channels, dtype, latency,
     return parameters, dtype, samplesize, samplerate
 
 
-def _wrap_callback(callback, data, frames, time, status):
+def _wrap_callback(callback, *args):
     """Invoke callback function and check for custom exceptions."""
     cf = CallbackFlags.__new__(CallbackFlags)
-    cf._flags = int(status)
+    cf._flags = int(args[-1])
+    args = args[:-1] + (cf,)
     try:
-        callback(data, frames, time, cf)
+        callback(*args)
     except CallbackStop:
         return _lib.paComplete
     except CallbackAbort:
--- a/src/sounddevice.py
+++ b/src/sounddevice.py
@@ -1927,7 +1927,7 @@ class CallbackFlags:
     __slots__ = '_flags'
 
     def __init__(self, flags=0x0):
-        self._flags = flags
+        self._flags = int(flags)
 
     def __repr__(self):
         flags = str(self)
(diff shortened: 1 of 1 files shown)


mgeier:
Thanks for this PR!

How can I reproduce the errors? Does this need a certain CFFI version?

IzaCarlos:
> How can I reproduce the errors? Does this need a certain CFFI version?

You can reproduce (I'm very sure) by installing Python v3.14.4 w/cffi 2.0.0 and Sounddevice with 0.5.5, then use any application that opens a stream for a *long* time.
Not to advertise or anything but one of my personal applications that monitors sound does use a continous stream and after being left idling it would eventually lead to this cffi error (pardon some mispells I had to OCR this since I couldn't copy paste from the box):

```
Exception ignored from cffi callback <function _StreamBase._init_.<locals>.callback_ptr at 0x000001DBC8B22140>:
File: "C:\Users\carlo\AppData\Local\Programs\Python\Python314\Li b\site-packages\sounddevice.py", line 862, in callback_ptr Exception ignored from cffi callback <function StreamBase. init <locals>.callback_ptr at 0x000001DBC8B22A30>:
File: "C:\Users\carlo\AppData\Local\Programs\Python\Python314\Lib\site-packages\sounddevice.py", line 878, in callback_ptr return_wrap_callback(callback, data, frames, time, status)
File: "C:\Users\carlo\AppData\Local\Programs\Python\Python314\Lib\site-packages\sounddevice.py", line 2775, in_wrap_callback return_wrap_callback(callback, data, frames, time, status)
File: "C:\Users\carlo\AppData\Local\Programs\Python\Python314\Li b\site-packages\sounddevice.py", line 2775, in_wrap_callback
args = tuple(args[:-1]) + (CallbackFlags(args[-1]),)

TypeError: __init__() should return None, not 'NoneType'
```

As mentioned I applied some fixes to my sd installation and it has not been causing any more issues.

mgeier:
I could not reproduce this on a Windows VM using any host API, so this bug might be more specific to your hardware.

I have the feeling your fix just hides an underlying problem, which should be fixed in CFFI or PortAudio (I don't know where the bug actually is).

Can you show the output of `python -m sounddevice`?

Can you try it with different host APIs?

Can you try it with different sound cards?

When the error appears, can you check which exact type the last argument in `_wrap_callback()` has?

You could try this by creating a file named `instrumented.py` in the same directory as your code:

```python
import faulthandler

import sounddevice as sd

faulthandler.enable()

_wrap_callback = sd._wrap_callback


def _instrumented_wrap_callback(callback, *args):
    status = args[-1]
    if type(status) is not int:
        print('ERROR: status is a', type(status))
        return sd._lib.paAbort
    try:
        str(sd.CallbackFlags(status))
    except BaseException as e:
        print('ERROR:', repr(e))
        return sd._lib.paAbort
    return _wrap_callback(callback, *args)


sd._wrap_callback = _instrumented_wrap_callback
```

... and import it at the very top or your code:

```python
import instrumented
... # the rest of your program, as you ran it before
```

Issue #460, opened 2023-03-12
Error opening OutputStream: Unanticipated host error [PaErrorCode -9999]


nscotto:
I am having this issue on a macOS 13.2.1 with M1 chip:
```shell
❯ python play_file.py MR1106_BlueLitMoon_Full/01_Kick.wav
||PaMacCore (AUHAL)|| AUHAL component not found.PortAudioError: Error opening OutputStream: Unanticipated host error [PaErrorCode -9999]: '' [<host API not found> error 0]
```
play_file.py:
```python
import argparse
import threading

import sounddevice as sd
import soundfile as sf


def int_or_str(text):
    """Helper function for argument parsing."""
    try:
        return int(text)
    except ValueError:
        return text


parser = argparse.ArgumentParser(add_help=False)
parser.add_argument(
    '-l', '--list-devices', action='store_true',
    help='show list of audio devices and exit')
args, remaining = parser.parse_known_args()
if args.list_devices:
    print(sd.query_devices())
    parser.exit(0)
parser = argparse.ArgumentParser(
    description=__doc__,
    formatter_class=argparse.RawDescriptionHelpFormatter,
    parents=[parser])
parser.add_argument(
    'filename', metavar='FILENAME',
    help='audio file to be played back')
parser.add_argument(
    '-d', '--device', type=int_or_str,
    help='output device (numeric ID or substring)')
args = parser.parse_args(remaining)

event = threading.Event()

try:
    data, fs = sf.read(args.filename, always_2d=True)

    current_frame = 0

    def callback(outdata, frames, time, status):
        global current_frame
        if status:
            print(status)
        chunksize = min(len(data) - current_frame, frames)
        outdata[:chunksize] = data[current_frame:current_frame + chunksize]
        if chunksize < frames:
            outdata[chunksize:] = 0
            raise sd.CallbackStop()
        current_frame += chunksize

    stream = sd.OutputStream(
        samplerate=fs, device=args.device, channels=data.shape[1],
        callback=callback, finished_callback=event.set)
    with stream:
        event.wait()  # Wait until playback is finished
except KeyboardInterrupt:
    parser.exit('\nInterrupted by user')
except Exception as e:
    parser.exit(type(e).__name__ + ': ' + str(e))
```
also I have:
```
>>> sd._libname
'/opt/homebrew/Caskroom/miniforge/base/envs/audio/bin/../lib/libportaudio.dylib
```
[...]

PR #631, opened 2026-05-14
Fix architecture detection on Windows ARM64

bwoodsend:
When running an AMD64 Python on a Windows ARM64 host, platform.machine() returns the host architecture instead of the process architecture, causing sounddevice to try and load the wrong DLL.

    >>> import sounddevice
      File "C:\Users\bagpuss\AppData\Local\Python\pythoncore-3.14-64\Lib\site-packages\sounddevice.py", line 72, in <module>
        raise OSError('PortAudio library not found')
    OSError: PortAudio library not found

    During handling of the above exception, another exception occurred:

      File "<python-input-1>", line 1, in <module>
        import sounddevice
      File "C:\Users\bagpuss\AppData\Local\Python\pythoncore-3.14-64\Lib\site-packages\sounddevice.py", line 91, in <module>
        _lib: ... = _ffi.dlopen(_libname)
                    ~~~~~~~~~~~^^^^^^^^^^
    OSError: cannot load library 'C:\Users\bagpuss\AppData\Local\Python\pythoncore-3.14-64\Lib\site-packages\_sounddevice_data\portaudio-binaries\libportaudioarm64.dll': error 0x7e

The PROCESSOR_ARCHITECTURE environment variable can be used to get process architecture. As far as I know, it should always be enough to only check that variable but I wasn't feeling brave enough to not leave the old detection as a fallback.

### DIFF
--- a/.github/workflows/sounddevice-data.yml
+++ b/.github/workflows/sounddevice-data.yml
@@ -16,6 +16,8 @@ jobs:
             arch: 'x86'
           - os: windows-11-arm
             arch: 'arm64'
+          - os: windows-11-arm
+            arch: 'x64'
     runs-on: ${{ matrix.os }}
     steps:
       - name: Set up Python
@@ -25,7 +27,7 @@ jobs:
           architecture: ${{ matrix.arch }}
       - name: Double-check Python version
         run: |
-          python --version
+          python --version --version
       - name: Clone Git repository (with submodules)
         uses: actions/checkout@v6
         with:
--- a/.github/workflows/sounddevice-data.yml
+++ b/.github/workflows/sounddevice-data.yml
@@ -38,6 +38,7 @@ jobs:
         run: |
           python -m pip install .
       - name: Import module
+        shell: bash
         run: |
           python -m sounddevice
           python -c "import sounddevice as sd; print(sd._libname)"
@@ -47,6 +48,7 @@ jobs:
           python -c "import sounddevice as sd; assert not any(a['name'] == 'ASIO' for a in sd.query_hostapis())"
       - name: Import module (using the ASIO DLL)
         if: startsWith(matrix.os, 'windows')
+        shell: bash
         env:
           SD_ENABLE_ASIO: 1
         run: |
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,6 @@
 import os
 import platform
+import sysconfig
 from setuptools import setup
 
 MACOSX_VERSIONS = '.'.join([
@@ -11,7 +12,8 @@ MACOSX_VERSIONS = '.'.join([
 system = os.environ.get('PYTHON_SOUNDDEVICE_PLATFORM', platform.system())
 machine = platform.machine().lower()
 architecture0 = os.environ.get('PYTHON_SOUNDDEVICE_ARCHITECTURE',
-    'arm64' if machine in ['arm64', 'aarch64'] else platform.architecture()[0])
+    'arm64' if 'arm64' in sysconfig.get_platform() or 'aarch64' in sysconfig.get_platform()
+    else platform.architecture()[0])
 
 if system == 'Darwin':
     libname = 'libportaudio.dylib'
--- a/src/sounddevice.py
+++ b/src/sounddevice.py
@@ -55,6 +55,7 @@ import contextlib as _contextlib
 import os as _os
 import platform as _platform
 import sys as _sys
+import sysconfig as _sysconfig
 from ctypes.util import find_library as _find_library
 from _sounddevice import ffi as _ffi
 
@@ -75,7 +76,7 @@ except OSError:
     if _platform.system() == 'Darwin':
         _libname = 'libportaudio.dylib'
     elif _platform.system() == 'Windows':
-        if _platform.machine().lower() in ('arm64', 'aarch64'):
+        if 'arm64' in _sysconfig.get_platform() or 'aarch64' in _sysconfig.get_platform():
             _platform_suffix = 'arm64'
         else:
             _platform_suffix = _platform.architecture()[0]


mgeier:
Thanks for this PR!

Is there a way to check in CI that this previously failed? We currently have a test here:
https://github.com/spatialaudio/python-sounddevice/blob/2634256b1c3f2ef9419f5535c493b74646f6692e/.github/workflows/sounddevice-data.yml#L17-L18

It would be great to add a failing test and then fix it.

BTW, see also https://github.com/bastibe/python-soundfile/pull/460 for related discussions.

mgeier:
> As far as I know, it should always be enough to only check that variable but I wasn't feeling brave enough to not leave the old detection as a fallback.

It might always be defined, but it might also have additional values?
Could it be wow64?

It seems to be hard to find reliable information on what is guaranteed to be defined on which systems.

It might be best to check only for the one you are interested in (AMD64) and leave all the rest to the fallback (until somebody else complains).

bwoodsend:
> Is there a way to check in CI that this previously failed? We currently have a test here:

Looks like setup-python handles AMD64 on ARM64 well.

[Here's](https://github.com/bwoodsend/python-sounddevice/actions/runs/25882911099/job/76067182909#step:6:1) the error going the other way. Presumably the platform detection in `setup.py` needs the same treatment...

Incidentally, you might want to switch that step to bash or use separate steps since powershell ignores the exit code for all but the last command. i.e. If all but the last check failed, it would be considered a pass.

> It might always be defined, but it might also have additional values?
> It might be best to check only for the one you are interested in (AMD64) and leave all the rest to the fallback (until somebody else complains).

That is pretty much the same code in the end. All I'd get rid of is the `ARM64` key in that dict since it gets that one right anyway. With the current change, an unrecognised `PROCESSOR_ARCHITECTURE` value should lead to the fallback path as before.
```bash
$ py -3-arm64 -c 'import os; print(os.environ["PROCESSOR_ARCHITECTURE"])'
ARM64
$ py -3 -c 'import os; print(os.environ["PROCESSOR_ARCHITECTURE"])'
AMD64
$ py -3-32 -c 'import os; print(os.environ["PROCESSOR_ARCHITECTURE"])'
x86
```

bwoodsend:
That should be CI happy and everything that #460 was trying to do done.

mgeier:
Thanks for the updates!

> That is pretty much the same code in the end. All I'd get rid of is the `ARM64` key in that dict since it gets that one right anyway.

That's what I'd like to understand ... so you could get rid of `ARM64` but not `x86`?

I would like to know which code is there because we need it and which code is there for no particular reason. I'm not doubting that it works, but I would like to optimize the code for human understanding and future maintenance.

bwoodsend:
> That's what I'd like to understand ... so you could get rid of ARM64 but not x86?

You can because there's not such thing as running an ARM64 process on AMD64 or x86.

How about just...
```diff
-        if _platform.machine().lower() in ('arm64', 'aarch64'):
+        if _platform.machine().lower() in ('arm64', 'aarch64') and _os.environ.get("PROCESSOR_ARCHITECTURE", "") not in ("x86", "AMD64"):
            _platform_suffix = 'arm64'
        else:
            _platform_suffix = _platform.architecture()[0]
```
...?

mgeier:
Yes, I think this would be better.

I've also just seen https://github.com/bastibe/python-soundfile/pull/489 which provides a nice overview table and suggests using `sysconfig.get_platform()` ... what do you think about that?

bwoodsend:
Hmm, when I did this yonks ago, I was put off `sysconfig.get_platform()` when I saw it was implemented via a ladder of [`'amd64' in sys.version.lower()`](https://github.com/python/cpython/blob/77b45fa6d0b8c0c14657b5117b21a3f3f2ce97d8/Lib/sysconfig/__init__.py#L574-L581) checks. It also produced different results for MinGW Python. ([Discussion](https://github.com/pyinstaller/pyinstaller/issues/8219#issuecomment-1885776570) in case you're interested.)

Now it's defined [based on compile time target detection](https://github.com/python/cpython/blob/f051c68923b4060b03566d0ea1df06d911ebe238/Modules/_sysconfig.c#L84-L117) (albeit [with a fallback to sys.platform](https://github.com/python/cpython/blob/f051c68923b4060b03566d0ea1df06d911ebe238/Lib/sysconfig/__init__.py#L673-L675)). It is also what pip uses ([via `packaging`](https://github.com/pypa/packaging/blob/b34d12acb28c9ad3a6b0b3cc82f03a4b0b98c8c0/src/packaging/tags.py#L793-L812)) so presumably it has to work no matter how it's done.

I'm pretty certain all 3 options will work so I'm good with whichever one sounds the least icky to you...

mgeier:
Thanks for the additional information. Everything related to Windows sounds icky to me!

But I think it would be good to avoid relying on environment variables if that's at all feasible (and it now seems to be).

bwoodsend:
How about...
```diff
diff --git i/src/sounddevice.py w/src/sounddevice.py
index 00fc6f8..ad38411 100644
--- i/src/sounddevice.py
+++ w/src/sounddevice.py
@@ -55,6 +55,7 @@ import contextlib as _contextlib
 import os as _os
 import platform as _platform
 import sys as _sys
+import sysconfig as _sysconfig
 from ctypes.util import find_library as _find_library
 from _sounddevice import ffi as _ffi
 
@@ -75,7 +76,7 @@ except OSError:
     if _platform.system() == 'Darwin':
         _libname = 'libportaudio.dylib'
     elif _platform.system() == 'Windows':
-        if _platform.machine().lower() in ('arm64', 'aarch64'):
+        if 'arm64' in _sysconfig.get_platform() or 'aarch64' in _sysconfig.get_platform():
             _platform_suffix = 'arm64'
         else:
             _platform_suffix = _platform.architecture()[0]
```
... then? The `or 'aarch64' in _sysconfig.get_platform()` can go away if you aren't interested in MINGW/MSYS2 (where `sysconfig.get_platform()` returns `mingw_aarch64_ucrt_llvm`).

mgeier:
Sorry for the late response ... this looks very good to me, nice and simple!

If we can support MINGW/MSYS2, that's great!

mgeier:
Thanks for following through with this!

